PR #23, opened 2015-01-20
npalamar / create background chat image and fix some ui bugs

npalamar:
Created backgound image for chats with corresponding logic
- Set navigation bar, tab bar translucent to NO;
- Fix some ui bugs when screen rotares;
- Now width depends on screen widht (width * coefficient)
- Remove code for hiding separator line in Frieds VC
- Add tableView footers with zeroFrame for removing unusefull lines

I have some problems with images and i guess i'll solve them this week


dvor:
It [doesn't build](https://travis-ci.org/dvor/Antidote/builds/47699653). It seems that you've removed all toxcore files in last commit bfbe2c1.

After some thoughts I decided that we don't need background images in Antidote (at least for now). Almost any image kills color scheme. Image doesn't fit bubbles and buttons color, messages became less readable. Really, it looks like a mess and I don't think it is a good user experience.

ios simulator screen shot 29 2015 23 42 05

----

However launch images and bug fixes are okay. Can you make another pull request with it?

*Thank you for your work. Please, don't make such complicated pull requests in the future. One thing at once would be much better.*

npalamar:
Hi! Yes I can next week. I'll split this branch into two another branches, first branch will contain bug fixes and launch images, second will contain chat background images functionality (may be in the future it will have sense).

PR #31, opened 2015-07-02
Start of audio implementation

Chuongv:
Still a lot of work to do but some progress. Would like some comments or suggestions to make sure I am on the right path :smile:


Antidote/CallViewController.h
```diff
@@ -0,0 +1,32 @@
...
+//  Created by Chuong Vu on 7/1/15.
+//  Copyright (c) 2015 dvor. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+@class OCTSubmanagerCalls;
+@class OCTChat;
+
+extern NSString *const kCallViewControllerUserIdentifier;
```
dvor: For what do you need this?

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+        return nil;
+    }
+
+    _chat = chat;
+
+    NSPredicate *predicate = [NSPredicate predicateWithFormat:@"chat == %@", chat];
+    _callController = [Helper createFetchedResultsControllerForType:OCTFetchRequestTypeCall predicate:predicate delegate:self];
+    _callController.delegate = self;
+
+    _call = [_callController fetchedObjects].firstObject;
```
dvor: `firstObject` is a method, not a property. In this cases bracket notation is preferred. `[[_callController fetchedObjects] firstObject]`.

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+- (instancetype)initWithChat:(OCTChat *)chat submanagerCalls:(OCTSubmanagerCalls *)manager
+{
+    self = [super init];
+    if (! self) {
+        return nil;
+    }
+
+    _chat = chat;
+
+    NSPredicate *predicate = [NSPredicate predicateWithFormat:@"chat == %@", chat];
```
dvor: For some reason RBQFetchedResultsController doesn't get updated with such predicate. In ChatViewController I had to do `chat.uniqueIdentifier == %@", ...`

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+#import "Helper.h"
+#import "OCTCall.h"
+#import "OCTChat.h"
+#import "OCTSubmanagerCalls.h"
+
+NSString *const kCallViewControllerUserIdentifier = @"callViewController";
+
+@interface CallViewController () <RBQFetchedResultsControllerDelegate>
+
+@property (strong, nonatomic) OCTSubmanagerCalls *manager;
```
dvor: I think that it would be better not to capture submanager. Controller doesn't control submanager's lifecycle at all. So `weak` would be more preferable.

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+@property (strong, nonatomic, readwrite) OCTChat *chat;
```
dvor: There is no need in storing chat, you can always get it from call

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+
+    self.view.opaque = YES;
+
+    [self setupBlurredView];
+    [self createNameLabel];
+
+    [self createEndCallButton];
+    [self createCallTimer];
+
+    [self layoutSubviews];
```
dvor: > You should not call this method directly. If you want to force a layout update, call the setNeedsLayout method instead to do so prior to the next drawing update. If you want to update the layout of your views immediately, call the layoutIfNeeded method.

[link](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIView_Class/#//apple_ref/occ/instm/UIView/layoutSubviews)

Also there is no need in asking for layout, it will happen automatically after viewDidLoad.

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+#pragma mark -  RBQFetchedResultsControllerDelegate
+
+- (void)controllerWillChangeContent:(RBQFetchedResultsController *)controller
+{}
+
+- (void)controller:(RBQFetchedResultsController *)controller didChangeObject:(RBQSafeRealmObject *)anObject atIndexPath:(NSIndexPath *)indexPath forChangeType:(NSFetchedResultsChangeType)type newIndexPath:(NSIndexPath *)newIndexPath
+{}
+
+- (void)controller:(RBQFetchedResultsController *)controller didChangeSection:(RBQFetchedResultsSectionInfo *)section atIndex:(NSUInteger)sectionIndex forChangeType:(NSFetchedResultsChangeType)type
+{}
```
dvor: No need for implementing this methods as they are optional.

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+
+- (void)controllerDidChangeContent:(RBQFetchedResultsController *)controller
+{
+    self.call = [self.callController fetchedObjects].firstObject;
```
dvor: `[[ ... ] firstObject];`

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+
+    self.timerLabel.text = [self convertToTimeFromInterval:self.call.callDuration];
+    [self.timerLabel setNeedsDisplay];
```
dvor: It would be better to move these two lines to separate method like `updateTimerLabel` and
- call it after configuring label, to have initial value
- call it here

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+}
+#pragma mark - Private
+
+- (void)setupBlurredView
+{
+    self.view.backgroundColor = [UIColor clearColor];
```
dvor: This isn't part of blurredView setup, you can move it to viewDidLoad

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+
+    UIView *darkView = [[UIView alloc] initWithFrame:self.view.bounds];
```
dvor: What will happen on rotation of device? darkView and blurEffect would have wrong frame?

There is no need to set frame in init method, you can just set constraints for both of them to `make.edges.equalTo(self.view);`

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+        make.left.equalTo(self.view.mas_left).with.offset(50);
+        make.right.equalTo(self.view.mas_right).with.offset(-50);
+    }];
+
+    [self.timerLabel mas_makeConstraints:^(MASConstraintMaker *make) {
+        make.top.equalTo(self.nameLabel.mas_bottom).with.offset(20);
+        make.left.equalTo(self.nameLabel.mas_left);
+        make.right.equalTo(self.nameLabel.mas_right);
+        make.centerX.equalTo(self.nameLabel.centerX);
+    }];
```
dvor: You don't want to make constraints every time layoutSubviews method is called. Please create method like `installConstraints` and call it at the end of viewDidLoad

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+
+    [self.view addSubview:visualEffectView];
+}
+
+- (void)createNameLabel
+{
+    if (self.call.chat.friends.count == 1) {
+        OCTFriend *friend = self.call.chat.friends.firstObject;
+
+        CGRect labelFrame = CGRectMake(0, 0, 200, 200);
```
dvor: Please use constraints for that.

Generally speaking, almost all Antidote code **doesn't** use auto layout and sets frames manually. However auto layout is better. That's why we want to use constraints **everywhere** in a new code (and migrate old code after time) :-)

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+        self.nameLabel.textColor = [UIColor whiteColor];
+        self.nameLabel.textAlignment = NSTextAlignmentCenter;
+        [self.nameLabel sizeToFit];
+
+        [self.view addSubview:self.nameLabel];
+    }
+}
+- (void)createEndCallButton
+{
+    CGRect frame = CGRectMake(0, 0, 400, 100);
```
dvor: Same for this frame

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+        make.left.equalTo(self.view.mas_left).with.offset(50);
+        make.right.equalTo(self.view.mas_right).with.offset(-50);
+        make.height.equalTo(30);
+    }];
+
+    [self.cancelCallButton mas_makeConstraints:^(MASConstraintMaker *make) {
+        make.bottom.equalTo(self.view.mas_bottom).with.offset(-50);
+        make.centerX.equalTo(self.view.mas_centerX);
+        make.left.equalTo(self.view.mas_left).with.offset(50);
+        make.right.equalTo(self.view.mas_right).with.offset(-50);
```
dvor: It would be nicer to replace magic numbers with constants (after imports), like
```
static const CGFloat kIndent = 50.0;
```

Antidote/CallViewController.m
```diff
@@ -0,0 +1,184 @@
...
+        make.centerX.equalTo(self.nameLabel.centerX);
+    }];
+}
+
+- (NSString *)convertToTimeFromInterval:(NSTimeInterval)interval
+{
+    int minutes = (int)interval / 60;
+    int seconds = interval - (minutes * 60);
+
+    return [NSString stringWithFormat:@"%02d:%02d", minutes, seconds];
```
dvor: I think there is TimeFormatter class for that

Antidote/ChatViewController.m
```diff
@@ -315,6 +318,14 @@ - (void)updateFriendRelatedInformation
     [self updateTitleView];
 }
 
+- (void)createPhoneCallBarButton
+{
+    UIImage *phoneImage = [UIImage imageNamed:@"phone"];
+    self.navigationItem.rightBarButtonItem = [[UIBarButtonItem alloc] initWithImage:phoneImage
+                                                                              style:UIBarButtonItemStylePlain
+                                                                             target:self
+                                                                             action:@selector(startCall)];
```
dvor: `startCallButtonPressed` would be better.

Antidote/Images.xcassets/Controllers/Call/phone.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "scale" : "1x"
+    },
+    {
+      "idiom" : "universal",
+      "scale" : "2x",
+      "filename" : "phone.png"
```
dvor: Please use more meaningful names for images (you can check other images). Also there should be `@2x` suffix for 2x images

 `call-phone@2x.png`

Antidote/ToxListener.m
```diff
@@ -124,6 +126,24 @@ - (void)controllerDidChangeContent:(RBQFetchedResultsController *)controller
...
 }
 
+#pragma mark - SubmanagerCalls delegate
+- (void)callSubmanager:(OCTSubmanagerCalls *)callSubmanager receiveCall:(OCTCall *)call audioEnabled:(BOOL)audioEnabled videoEnabled:(BOOL)videoEnabled
+{
+    AppDelegate *delegate = (AppDelegate *)[UIApplication sharedApplication].delegate;
+    UIViewController *visibleVC = [delegate visibleViewController];
+
+    if ([visibleVC isKindOfClass:[CallViewController class]]) {
+        // User is in a middle of call, send some kind of notification?
```
dvor: Is it possible in current toxav to have incoming call while being in call? Can we have several calls and switch between them?

Antidote/ToxListener.m
```diff
@@ -124,6 +126,24 @@ - (void)controllerDidChangeContent:(RBQFetchedResultsController *)controller
...
+        [callSubmanager sendCallControl:OCTToxAVCallControlCancel toCall:call error:nil];
+        return;
+    }
+
+    CallViewController *callViewController = [[CallViewController alloc] initWithChat:call.chat submanagerCalls:callSubmanager];
+    callViewController.modalInPopover = YES;
+    callViewController.modalPresentationStyle = UIModalPresentationOverFullScreen;
+    [visibleVC presentViewController:callViewController animated:YES completion:nil];
```
dvor: You didn't send answer control

dvor:
This looks good so far. However I'm thinking now that it would be better to split CallViewController to several controllers. So what we can create is
- Ringing controller. It will be responsible for showing incoming caller name, playing music/vibrating, will have accept/decline buttons
- Dialing controller. It will be responsible for showing caller name, playing ring-back tones, will have decline button
- Active call controller.

When user press call button you can
- create dialing controller with chat
- put it into navigation controller and present
- dialing controller will be responsible for actually calling objcTox methods and starting call
- dialing controller will push without animation active controller

Same for ringing controller. You can create it with incoming call, present, and then controller will be responsible for answering call and doing whatever it wants.

What do you think about that? I hope I'll be able to provide you mockups for that soon.

Antidote/CallViewController.m
Chuongv: There's stuff for creating NSDate objects but I couldn't find anything for converting time interval to MM:SS. There was NSNumberFormatter and NSDateFormatter, but those don't seem to fit my need.

Chuongv:
Yea I was debating whether or not to break the CallViewController into several or just have one and animate between the states depending on the OCTCallStatus. But I think your way sounds better!

Antidote/ToxListener.m
Chuongv: Mannol said yes to both questions. I guess then we need to change that in objcTox as well then. Though it shouldn't be that big of a change. Will need to find a way to get a third device to test this :sweat_smile:

dvor: Well, I have enough devices, don't worry. :-)

Antidote/AbstractCallViewController.h
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * The call associated with this view.
+ */
+@property (strong, nonatomic) OCTCall *call;
+
+/**
+ * The call manager responsible for call handling.
+ */
+@property (weak, nonatomic) OCTSubmanagerCalls *manager;
```
dvor: No need for readwrite properties in public interface. You can add `readonly` here and leave `readwrite` in .m file

Antidote/AbstractCallViewController.h
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Create an instance of the CallViewController.
+ * @param chat Appropriate chat for the call.
+ * @param manager The call maanger for the application.
+ */
+- (instancetype)initWithChat:(OCTChat *)chat submanagerCalls:(OCTSubmanagerCalls *)manager;
+
+/**
+ * The call associated with this view.
```
dvor: with this *controller*.

dvor: Also you can mention that it is autoupdating call. It would be nice to have public method `- (void)didUpdateCall NS_REQUIRES_SUPER;`
- which will be called by abstract controller after updating call
- abstract controller can do stuff some stuff in this method if it like
- subclass can override it if it like and do some other stuff.

Antidote/AbstractCallViewController.m
```diff
@@ -0,0 +1,82 @@
...
+#pragma mark - Life cycle
+
+- (instancetype)initWithChat:(OCTChat *)chat submanagerCalls:(OCTSubmanagerCalls *)manager
+{
+    self = [super init];
+    if (! self) {
+        return nil;
+    }
+
+    NSPredicate *predicate = [NSPredicate predicateWithFormat:@"chat == %@", chat];
```
dvor: `@"chat.uniqueIdentifier == %@", chat.uniqueIdentifier`

Antidote/AbstractCallViewController.m
```diff
@@ -0,0 +1,82 @@
...
+    _callController = [Helper createFetchedResultsControllerForType:OCTFetchRequestTypeCall predicate:predicate delegate:self];
+    _callController.delegate = self;
+
+    _call = [[_callController fetchedObjects] firstObject];
+
+    if (! _call) {
```
dvor: There may be no call in init method. For example Dialing controller will get call only when it will start calling in chat.

Antidote/AbstractCallViewController.m
```diff
@@ -0,0 +1,82 @@
...
+
+- (void)setupBlurredView
+{
+    UIView *darkView = [[UIView alloc] initWithFrame:self.view.bounds];
+    darkView.backgroundColor = [UIColor blackColor];
+    darkView.alpha = 0.5;
+
+    [self.view addSubview:darkView];
+
+    [darkView mas_makeConstraints:^(MASConstraintMaker *make) {
```
dvor: btw, you can use Masonry's methods without `mas_` prefix. Just `makeConstraints:`

Antidote/ChatViewController.m
```diff
@@ -373,4 +384,21 @@ - (void)updateLastReadDate
     [[AppContext sharedContext].profileManager.toxManager.objects changeChat:self.chat lastReadDateInterval:interval];
 }
 
+- (void)startCallButtonPressed
+{
+    OCTSubmanagerCalls *manager = [AppContext sharedContext].profileManager.toxManager.calls;
+
+    OCTCall *call = [manager callToChat:self.chat enableAudio:YES enableVideo:NO error:nil];
```
dvor: So this should go to Dialing controller

Antidote/ChatViewController.m
```diff
@@ -373,4 +384,21 @@ - (void)updateLastReadDate
...
+
+    if (! call) {
+        return;
+    }
+
+    DialingCallViewController *dailingCallViewController = [[DialingCallViewController alloc] initWithChat:self.chat submanagerCalls:manager];
+
+    dailingCallViewController.modalInPopover = YES;
+    dailingCallViewController.modalPresentationStyle = UIModalPresentationOverFullScreen;
+    [self presentViewController:dailingCallViewController animated:YES completion:nil];
```
dvor: Dialing controller should be put inside navigation (with hidden navigation bar). That way it would be able to push Active controller

Antidote/DialingCallViewController.m
```diff
@@ -0,0 +1,159 @@
...
+
+    [self createEndCallButton];
+    [self createCallTimer];
+
+    [self installConstraints];
+}
+
+#pragma mark -  RBQFetchedResultsControllerDelegate
+
+- (void)controllerDidChangeContent:(RBQFetchedResultsController *)controller
```
dvor: It would be better to move this method into Abstract controller and add method `didUpdateCall` I was talking about.

Antidote/DialingCallViewController.m
```diff
@@ -0,0 +1,159 @@
...
+
+    UIVisualEffect *blurEffect = [UIBlurEffect effectWithStyle:UIBlurEffectStyleLight];
+
+    UIVisualEffectView *visualEffectView = [[UIVisualEffectView alloc] initWithEffect:blurEffect];
+    visualEffectView.frame = self.view.bounds;
+
+    [self.view addSubview:visualEffectView];
+    [visualEffectView mas_makeConstraints:^(MASConstraintMaker *make) {
+        make.edges.equalTo(self.view);
+    }];
```
dvor: Dublicate?

Antidote/AbstractCallViewController.m
Chuongv: So you are talking about calling straight from the friends profile? Right now this initializer requires a chat instance. My intention was that a valid chat instance must be available before any calls incoming/outgoing are made. Not sure if there is a better way to do this?

dvor: Hm. Subclass can be responsible for creating/getting call.

Also Ringing controller will be created with existing call. It can just pass it to superclass. Pseudocode:

```
// Abstract controller
- (instancetype)initWithCall:(OCTCall *)call submanagerCalls:...
{
   ...
   _call = call;
   // get call.chat.uniqueIdentifier and subsribe for updates
}

// Dialing controller
- (instancetype)initWithChat:(OCTChat *)chat submanager...
{
    call = [submanager getCallFromChat:chat];
    self = [super initWithCall:call submanager...
...

// Ringing controller
- (instancetype)initWithCall:(OCTCall *)call submanagerCalls:...
{
    self = [super initWithCall:call submanager...
...
```

dvor: In subclasses you can duplicate Abstract's method with 'NS_UNAVAILABLE` suffix so no one will be able to use it for creating subclasses.

```
- (instancetype)initWithCall:(OCTCall *)call submanagerCalls:(..)calls NS_UNAVAILABLE;
```

Antidote/ChatViewController.m
Chuongv: If I do this, what's the best way to give the user an option to call with no audio? Or do you want them to have that selection in the `DialingCallViewController` before calling?

Chuongv: Well come to think of it, I don't think any user would do that. So nevermind

dvor: Well, similarly to regular phone, Skype, Viber, whatever, user will always call with audio by default. If she likes she can disable audio right after start of dialing.

Chuongv:
Sorry, got this PR request a little too big. I won't push anymore more new features, unless it's a correction.

Antidote/AbstractCallViewController.h
dvor: readonly for manager?

Antidote/AbstractCallViewController.m
dvor: Still `mas_` prefix here

Antidote/AbstractCallViewController.m
```diff
@@ -0,0 +1,131 @@
...
+    }
+}
+
+- (void)installConstraints
+{
+    [self.nameLabel makeConstraints:^(MASConstraintMaker *make) {
+        make.top.equalTo(self.view.mas_top).with.offset(kIndent);
+        make.left.equalTo(self.view.mas_left).with.offset(kIndent);
+        make.right.equalTo(self.view.mas_right).with.offset(-kIndent);
+        make.centerX.equalTo(self.view.mas_centerX);
```
dvor: Hm, no need in `left` and `right` in case if you set `centerX`. Also no need in `mas_` prefix here as well.

Antidote/ActiveCallViewController.m
```diff
@@ -0,0 +1,103 @@
...
+
+    [self installConstraints];
+}
+
+- (void)createEndCallButton
+{
+    self.endCallButton = [UIButton new];
+    self.endCallButton.backgroundColor = [UIColor redColor];
+    [self.endCallButton setImage:[UIImage imageWithContentsOfFile:@"phone"] forState:UIControlStateNormal];
+    [self.endCallButton addTarget:self action:@selector(endCall) forControlEvents:UIControlEventTouchUpInside];
```
dvor: Please rename `endCall` to `endCallButtonPressed`

Antidote/ActiveCallViewController.m
```diff
@@ -0,0 +1,103 @@
...
+        make.left.equalTo(self.nameLabel.mas_left);
+        make.right.equalTo(self.nameLabel.mas_right);
+        make.centerX.equalTo(self.nameLabel.centerX);
+    }];
+
+    [self.endCallButton makeConstraints:^(MASConstraintMaker *make) {
+        make.bottom.equalTo(self.view.mas_bottom).with.offset(-kIndent);
+        make.centerX.equalTo(self.view.mas_centerX);
+        make.left.equalTo(self.view.mas_left).with.offset(kIndent);
+        make.right.equalTo(self.view.mas_right).with.offset(-kIndent);
```
dvor: No need for `mas_` prefix. You can do search for `mas_` and remove it everywhere.

dvor: And again, here you have `centerX`, no need for `left` and `right`. Same applies to `timerLabel`

Antidote/ActiveCallViewController.m
```diff
@@ -0,0 +1,103 @@
...
+    [self.manager sendCallControl:OCTToxAVCallControlCancel toCall:self.call error:nil];
+}
+
+- (void)updateTimerLabel
+{
+    self.timerLabel.text = [self stringFromTimeInterval:self.call.callDuration];
+    [self.timerLabel setNeedsDisplay];
+}
+
+- (NSString *)stringFromTimeInterval:(NSTimeInterval)interval
```
dvor: So instead of this method you can create NSDate (`dateWithTimeIntervalSince1970:`) and pass it to `TimeFormatter`. If needed format is missing you can add it there.

Actually OCTCall should have method returning `NSDate`. For now you can create it manually, but add a comment mentioning this [issue](https://github.com/Antidote-for-Tox/objcTox/issues/55).

Antidote/DialingCallViewController.m
```diff
@@ -0,0 +1,102 @@
...
+
+#pragma mark - Private
+
+
+- (void)createEndCallButton
+{
+    self.cancelCallButton = [UIButton new];
+    self.cancelCallButton.backgroundColor = [UIColor redColor];
+    [self.cancelCallButton setImage:[UIImage imageWithContentsOfFile:@"phone"] forState:UIControlStateNormal];
+    [self.cancelCallButton addTarget:self action:@selector(endCall) forControlEvents:UIControlEventTouchUpInside];
```
dvor: Please rename `endCall` to `endCallButtonPressed`

Antidote/DialingCallViewController.m
```diff
@@ -0,0 +1,102 @@
...
+
+- (void)installConstraints
+{
+    [super installConstraints];
+
+    [self.cancelCallButton makeConstraints:^(MASConstraintMaker *make) {
+        make.bottom.equalTo(self.view.mas_bottom).with.offset(-kIndent);
+        make.centerX.equalTo(self.view.mas_centerX);
+        make.left.equalTo(self.view.mas_left).with.offset(kIndent);
+        make.right.equalTo(self.view.mas_right).with.offset(-kIndent);
```
dvor: No need for `left`/`right`, remove `mas_`

Antidote/RingingCallViewController.m
```diff
@@ -0,0 +1,84 @@
...
+//  Copyright (c) 2015 dvor. All rights reserved.
+//
+
+#import "RingingCallViewController.h"
+#import "OCTSubmanagerCalls.h"
+#import "ActiveCallViewController.h"
+#import "Masonry.h"
+
+static const CGFloat kIndent = 50.0;
+static const CGFloat kButtonsize = 50.0;
```
dvor: kButton**S**ize

Antidote/DialingCallViewController.m
```diff
@@ -0,0 +1,102 @@
...
+}
+
+#pragma mark - Private
+
+
+- (void)createEndCallButton
+{
+    self.cancelCallButton = [UIButton new];
+    self.cancelCallButton.backgroundColor = [UIColor redColor];
+    [self.cancelCallButton setImage:[UIImage imageWithContentsOfFile:@"phone"] forState:UIControlStateNormal];
```
dvor: You don't have `phone` image anymore

Antidote/ActiveCallViewController.m
```diff
@@ -0,0 +1,103 @@
...
+    [self createCallTimer];
+
+    [self installConstraints];
+}
+
+- (void)createEndCallButton
+{
+    self.endCallButton = [UIButton new];
+    self.endCallButton.backgroundColor = [UIColor redColor];
+    [self.endCallButton setImage:[UIImage imageWithContentsOfFile:@"phone"] forState:UIControlStateNormal];
```
dvor: You don't have `phone` image anymore

Antidote/ActiveCallViewController.m
```diff
@@ -0,0 +1,100 @@
...
+- (NSString *)stringFromTimeInterval:(NSTimeInterval)interval
+{
+    // See https://github.com/Antidote-for-Tox/objcTox/issues/55
+    NSDate *date = [NSDate dateWithTimeIntervalSince1970:interval];
+
+    NSDateFormatter *dateFormatter = [NSDateFormatter new];
+    [dateFormatter setTimeZone:[NSTimeZone timeZoneForSecondsFromGMT:0]];
+    [dateFormatter setDateFormat:@"HH:mm:ss"];
+
+    return [dateFormatter stringFromDate:date];
```
dvor: So you can revert this back, everything was cool. %)
Also AFAIU no need in DateFormatter

PR #36, opened 2015-07-06
Enable/Disable call button based on friend's online status

Chuongv:
Small PR for enable/disabling call button based on the status of friend.

### DIFF
--- a/Antidote/ChatViewController.m
+++ b/Antidote/ChatViewController.m
@@ -91,9 +91,9 @@ NSString *const kChatViewControllerUserIdentifier = @"user";
 
     [self configureInputToolbar];
 
-    [self updateFriendRelatedInformation];
-
     [self createPhoneCallBarButton];
+
+    [self updateFriendRelatedInformation];
 }
 
 - (void)viewWillAppear:(BOOL)animated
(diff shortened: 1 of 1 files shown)


Antidote/ChatViewController.m
```diff
@@ -321,10 +321,15 @@ - (void)updateFriendRelatedInformation
...
     UIImage *phoneImage = [UIImage imageNamed:@"call-phone"];
+
     self.navigationItem.rightBarButtonItem = [[UIBarButtonItem alloc] initWithImage:phoneImage
                                                                               style:UIBarButtonItemStylePlain
                                                                              target:self
                                                                              action:@selector(startCallButtonPressed)];
+
+    if (self.friend.connectionStatus == OCTToxConnectionStatusNone) {
+        self.navigationItem.rightBarButtonItem.enabled = NO;
+    }
```
dvor: Instead of this it would be better to move `createPhoneCallBarButton` before `updateFriendRelatedInformation` in `viewDidLoad` method.

Chuongv: Oh yea, good idea

PR #38, opened 2015-07-07
Add background mode voip and audio in project settings

Chuongv:
Haven't really tested this yet if this keeps the call alive in the background. But this is needed if we later decide to keep a connection to toxav for incoming calls.


dvor:
Can you check if it works? :-)

Chuongv:
Just tested, the call stays alive after the screen locks :+1:

dvor:
Cool :+1:

PR #39, opened 2015-07-07
Improve blurred background

Chuongv:
Moved the navigation controller to it's own class, along with the
blurred logic.

screen shot 2015-07-07 at 3 22 22 pm

Might be a little too dark? I could make the alpha `0.9` as below:

screen shot 2015-07-07 at 3 25 39 pm


dvor:
I think it is okay for now, we can always adjust it in future.

PR #40, opened 2015-07-08
Add appearance manager and implement avatar for call screens

Chuongv:
Would like your thoughts on this. I'm not sure if I should have created an extra method for the avatar method? Also would it be better to pass the `AvatarManager` as a property to the `AbstractCallViewController`?


Antidote/AbstractCallViewController.h
```diff
@@ -34,6 +36,8 @@
  */
 @property (weak, nonatomic) OCTSubmanagerCalls *manager;
 
+@property (weak, nonatomic) AppearanceManager *appearanceManager;
```
dvor: I think that you can remove this property and use `appearance` directly form AppContext. That's way it is used all over the project.

Same for `avatars`.

Antidote/AvatarsManager.m
```diff
@@ -98,6 +98,44 @@ - (UIImage *)createAvatarFromString:(NSString *)string diameter:(CGFloat)diamete
     return [self imageWithView:label];
 }
 
+- (UIImage *)createAvatarFromString:(NSString *)string diameter:(CGFloat)diameter
+                          textColor:(UIColor *)textColor backgroundColor:(UIColor *)backgroundColor
```
dvor: To avoid code duplication you should refactor `createAvatarFromString:diameter:` to call this method with colors from `appearance`

Antidote/DialingCallViewController.m
```diff
@@ -61,10 +63,30 @@ - (void)createEndCallButton
     [self.view addSubview:self.cancelCallButton];
 }
 
+- (void)createFriendAvatar
+{
+    AvatarsManager *avatars = [[AvatarsManager alloc] init];
+
+    CGFloat diameter = (1.0 / 2.0) * self.view.bounds.size.width;
```
dvor: This will be broken in landscape. I think you can just define some constant for diameter, that would be enought.

Antidote/DialingCallViewController.m
```diff
@@ -61,10 +63,30 @@ - (void)createEndCallButton
     [self.view addSubview:self.cancelCallButton];
 }
 
+- (void)createFriendAvatar
+{
+    AvatarsManager *avatars = [[AvatarsManager alloc] init];
```
dvor: Here you want to use `avatars` from AppContext

Antidote/DialingCallViewController.m
```diff
@@ -61,10 +63,30 @@ - (void)createEndCallButton
...
+
+    CGFloat diameter = (1.0 / 2.0) * self.view.bounds.size.width;
+
+    OCTFriend *friend = [self.call.chat.friends firstObject];
+    UIImage *image = [avatars createAvatarFromString:friend.name diameter:diameter
```
dvor: You should use `nickname` instead of `name`, as it is used all over the project.

Antidote/RingingCallViewController.m
```diff
@@ -41,9 +44,26 @@ - (void)viewDidLoad
...
+{
+    AvatarsManager *avatars = [[AvatarsManager alloc] init];
+
+    CGFloat diameter = (1.0 / 2.0) * self.view.bounds.size.width;
+
+    OCTFriend *friend = [self.call.chat.friends firstObject];
+    UIImage *image = [avatars createAvatarFromString:friend.name diameter:diameter
+                                           textColor:[UIColor whiteColor] backgroundColor:[UIColor clearColor]];
+
+    self.friendAvatar = [[UIImageView alloc] initWithImage:image];
```
dvor: All comments from DialingVC applies here too.

Antidote/AbstractCallViewController.h
Chuongv: I was trying to avoid singletons, but then I just realized `Appcontext` is imported everywhere, so I guess I should just use it then.

Antidote/AvatarsManager.m
Chuongv: I am not sure what you mean? There is already one using the color from `appearance`. However I am trying to get the white instead of the app's tint color. 

Ohh, I think you mean using the other colors available from the `AppearanceManager`, like `- (UIColor *)textMainColor;`?

dvor: No, I meant refactoring previous method to something like that:

```
- (UIImage *)createAvatarFromString:(NSString *)string diameter:(CGFloat)diameter
{
    return [self createAvatarFromString:string
diameter:diameter
textColor:appearance.color
backgroundColor:appearance.backgroundColor];
}
```

Chuongv: Ohh okay now I get what you are saying :smile:

Antidote/AbstractCallViewController.h
```diff
@@ -12,6 +12,8 @@
 @class OCTChat;
 @class OCTCall;
 
+#import "AppearanceManager.h"
```
dvor: No need to add this import in header, you can move it to implementation file.

dvor:
See https://github.com/Antidote-for-Tox/Antidote/pull/41

Chuongv:
Okay so now I have to merge audio in here, resolve conflicts and push again

PR #41, opened 2015-07-08
Fixing public method in AvatarsManager

dvor:
You created wrong method, without caching. It was faster to fix it than explain what is wrong. :-)

Feel free to merge it.


Chuongv:
Thanks :+1:

PR #43, opened 2015-07-10
WIP multipleCallHandling

Chuongv:
Merely just a work in progress, but I want to make sure I have the right architecture before I invest any further. No need to merge yet.

The whole idea is to have the `CallNavigationViewController` handle all the business of moving all the `AbstractCallViewControllers` including notifying any incoming calls.

Comments, suggestions or critiques appreciated :smile:


Antidote/AbstractCallViewController.m
```diff
@@ -110,4 +110,14 @@ - (void)endCall
...
 }
+
+- (void)switchToCall:(OCTCall *)call
+{
+    [(CallNavigationViewController *)self.navigationController switchToCall:call fromAbstractViewController:self];
+}
+
+- (void)displayNotificationOfNewCall:(OCTCall *)call
+{
+    // To Do: Display incoming call.
```
dvor: Shouldn't comment say something like `// Implement in successors`?

dvor:
This look quite good.

However I think that even better it would be to move all logic out of controllers. You can create CallsManager:
- it will be property of AppContext and be lazy initialized
- it will have two methods
  - callToChat:
  - handleIncomingCall:
- from ChatViewController you will call first method, from ToxListener the second one
- ChatVC and ToxListener doesn't care if there is no call, there is active call or something else
- CallsManager will manage all calls
- CallsManager will present/dismiss push/pop controllers
- controllers will know nothing about current call state. They
  - will just have set of properties to show (like callerName, callDurationString, etc.). These properties will be set and updated by CallsManager
  - will have a delegate (set to CallsManager) and send a messages to it like "Dismiss button has been pressed", etc.
- when last call will end CallsManager can tell AppContext to kill itself ([[AppContext sharedContext] killCallsManager]).

If you have any other ideas we can discuss them. :-)

Antidote/AbstractCallViewController.m
Chuongv: Yea that's true

Chuongv:
Thanks! The above sounds a lot better than how I had it originally planned.

Chuongv:
So in this case, you are proposing to remove `RBQFetchedResultsControllerDelegate` from `AbstractCallsViewController`, into `CallsManager`? And each time a call gets updated, `CallsManager` will find the appropriate `AbstractViewController` and update the interface?

dvor:
Yes, CallsManager can hold reference to active controller and update it. Also it will push another controller when needed.

Chuongv:
Are we removing the `OCTSubmanagerCalls` from `AbstractCallViewController` as well? So I guess in this case, everything related to calls is going to go through `CallsManager`. The only thing `AbstractCallViewController` will have is a property for one `OCTCall` which `CallsManager` will use to uniquely identify.

dvor:
Why does CallsManager need to identifier controller? Manager knows which call is active.

dvor:
Btw, AppContext now has `tabBarController` property. You can use it to present calls controller from manager.

Chuongv:
>Why does CallsManager need to identifier controller? Manager knows which call is active.

Hmm true

>Btw, AppContext now has tabBarController property. You can use it to present calls controller from manager.

Awesome! That should be helpful

Chuongv:
>Why does CallsManager need to identifier controller? Manager knows which call is active.

If we are setting `CallsManager` as the delegate and removing `OCTSubmanagerCalls`from the `AbstractCallViewController`, we need to know which call to dismiss, so that `CallsManager` can send the call control the appropriate call. 

Otherwise we can leave `OCTCall` and `OCTSubmanagerCalls` with the `AbstractCallsViewController`. Not sure which is best.

dvor:
CallsManager can simply hold reference to active call.

There is only one active call and one active controller at the moment. CallsManager holds them both. :-)

Chuongv:
Okay, I think I followed most of what you suggested :smile: . Still a lot of work to do. Comments or suggestions are welcomed.

Chuongv:
Some notes about what I did and what else needs to be implemented.

CallsManager has been added.

The changing from Dialing/Ringing -> Active is triggered by monitoring for inserts of `OCTCalls` with status `OCTCallStatusActive`. If there is ever a delete and we end up with no calls, we then kill `CallsManager`.

To Do:
If call from Alice is in paused/ringing state, and call with Bob ends, we should automatically switch over to Alice's call screen. 
Create call notification for incoming calls.

Antidote/CallsManager.h
```diff
@@ -0,0 +1,35 @@
...
+//  Copyright (c) 2015 dvor. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@class OCTChat;
+@class OCTCall;
+#import "AbstractCallViewController.h"
+
+@interface CallsManager : NSObject <AbstractCallViewControllerDelegate>
```
dvor: No need to make this public. `#import "AbstractCallViewController.h"` can be moved to implementation as well.

Antidote/CallsManager.h
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Set the property to YES to enable the microphone, otherwise NO.
+ * Default value is YES;
+ **/
+@property (nonatomic, assign) BOOL enableMicrophone;
```
dvor: This property is part of protocol. No need to make it public as well.

Antidote/NSString+Utilities.h
```diff
@@ -15,4 +15,6 @@
 
 - (NSString *)substringToByteLength:(NSUInteger)length usingEncoding:(NSStringEncoding)encoding;
 
++ (NSString *)stringFromTimeInterval:(NSTimeInterval)interval;
```
dvor: Please add a documentation with string format `"01:23"`

Antidote/AbstractCallViewController.h
```diff
@@ -8,31 +8,45 @@
...
+@required
+- (void)dismissCurrentCall;
+- (void)callAccept;
+- (BOOL)sendCallControl:(OCTToxAVCallControl)control error:(NSError **)error;
+
+- (void)otherCallAccept:(BOOL)accept;
+
+@property (nonatomic, assign) BOOL enableMicrophone;
+
+@end
```
dvor: I have some ideas about improving this protocol. Don't have time now, will write you in a 2-3 hours. :-)

dvor: I think that it would be nice if controllers would be as light as possible. There is no need for controllers to know something about call. Also there is no need for them to know about OCTToxAVCallControl. They will be responsible only for presenting text and handling button touches.

You can pass touches and handle them in CallsManager. Controllers will have properties which CallsManager will update. In setter of property you can update UI immediately. You don't have to recreate controller when changing active call - just update caller name and call duration.

 Also I don't think that you should move everything to Abstract controller. You can leave it just for blurred background.

So API as I see it is following. I may be missing something, but hope you get general idea:

```
@class DialingCallViewController
@property (weak, nonatomic) id<DialingCallViewControllerDelegate> delegate;
@property (strong, nonatomic) NSString *callerName;
@end

@protocol DialingCallViewControllerDelegate
- (void)dialingCallDeclineButtonPressed;
@end
```

```
@class RingingCallViewController
@property (weak, nonatomic) id<RingingCallViewControllerDelegate> delegate;
@property (strong, nonatomic) NSString *callerName;
@end

@protocol RingingCallViewControllerDelegate
- (void)ringingCallAnswerButtonPressed;
- (void)ringingCallDeclineButtonPressed;
@end
```

```
@class ActiveCallViewController
@property (weak, nonatomic) id<AbstractCallViewControllerDelegate> delegate;

@property (strong, nonatomic) NSString *callerName;
@property (assign, nonatomic) NSTimeInterval callDuration;
@property (assign, nonatomic) BOOL muteSelected;
@property (assign, nonatomic) BOOL speakerSelected;

@property (assign, nonatomic) BOOL showIncomingCallView;
@property (strong, nonatomic) NSString *incomingCallCallerName;

// An array with "paused calls" objects. Each object has following properties:
// NSString *callerName;
// NSTimeInterval startHoldInterval; // from that time you can calculate on hold duration
@property (strong, nonatomic) NSArray *pausedCallsArray;
@end

@protocol ActiveCallViewControllerDelegate
- (void)activeCallDeclineButtonPressed;
- (void)activeCallMuteButtonPressed;
- (void)activeCallSpeakerButtonPressed;
- (void)activeCallDeclineIncomingCallButtonPressed;
- (void)activeCallAnswerIncomingCallButtonPressed;
- (void)activeCallPausedCallSelectedAtIndex:(NSUInteger)index;
@end
```

dvor: So, for example, when user pressed `mute` button following happens:
- controllers sends `activeCallMuteButtonPressed` message to CallsManager
- CallsManager sends mute or unmute control to objcTox
- CallsManager sets `muteSelected` property of controller to new value
- controller in setter `- (void)setMuteSelected:(BOOL)selected` updates UI.

Here is the flow for another incoming call:
- we have active call
- CallsManager gets another incoming call
- CallsManager
  - sets `showIncomingCallView` to `YES`
  - sets `incomingCallCallerName` to caller's name
- ...
- user taps answer button
- controllers sends `activeCallAnswerIncomingCallButtonPressed` to CallsManager
- CallsManager changes active call
- CallsManager sets `showIncomingCallView` to `NO`
- CallsManager sets `callerName` to new caller's name and updates `callDuration`
- CallsManager sets `pausedCallsArray` to new array with old call object
- controller in `- (void)setPausedCallsArray:` setter checks if `pausedCallsArray` is empty
- it isn't empty so controller shows top right button with badge (1)
- when clicking on top right button controller shows paused table with old call object.

Chuongv: hmm okay I think I got the idea. Will try and implement :smile:  Thanks!

Chuongv: Do you think it would be nicer to have it send itself as a reference as well in the protocol?

So instead of 
`- (void)activeCallDeclineButtonPressed;`

We could do

`- (void)activeCallDeclineButtonPressed:(ActiveCallViewController *)controller;`

dvor: Yeah, it would be more convenient.

Chuongv:
Okay with the latest changes, I got it to work so now we can pause and answer another call :telephone_receiver: 

However I am not sure how the transition should work if the Call B ends, while Call A is waiting on the background. Should we dismiss Call B's view controller and automatically switch to Call A? 

I still need to implement the pause array. Though I can just grab all paused calls from `RBQFetchController` right?

Antidote/AbstractCallViewController.h
```diff
@@ -18,21 +16,20 @@
...
  */
-- (instancetype)initWithCall:(OCTCall *)call submanagerCalls:(OCTSubmanagerCalls *)manager;
+- (instancetype)initWithCallerNickname:(NSString *)nickname;
 
 /**
- * The call associated with this view controller.
+ * Abstract delegate
  */
-@property (strong, nonatomic, readonly) OCTCall *call;
+@property (weak, nonatomic) id delegate;
```
dvor: Do we need this delegate without protocol? I think you can remove it and leave only concrete delegates of controllers.

Antidote/AbstractCallViewController.h
```diff
@@ -18,21 +16,20 @@
...
 
 /**
- * The call manager responsible for call handling.
+ * Name of the caller
  */
-@property (weak, nonatomic) OCTSubmanagerCalls *manager;
+@property (strong, nonatomic, readonly) NSString *nickname;
```
dvor: Aren't you planning to change `nickname` of controller when switching call? In this case there is no need in `readonly`, as well as in `initWithCallerNickname` method. You can create controller and then set `nickname` property

Antidote/ActiveCallViewController.m
```diff
@@ -18,6 +18,9 @@
 static const CGFloat kEndCallButtonHeight = 45.0;
 static const CGFloat kButtonBorderWidth = 1.5f;
 static const CGFloat k3ButtonGap = 30.0;
+static const CGFloat kotherCallButtonSize = 4.0 / 5.0 * kButtonSide;
```
dvor: k **O** therCallButtonSize

Antidote/ActiveCallViewController.m
```diff
@@ -26,12 +29,15 @@ @interface ActiveCallViewController ()
...
 @property (strong, nonatomic) UIButton *microphoneButton;
-@property (strong, nonatomic) UIButton *muteButton;
+@property (strong, nonatomic) UIButton *speakerButton;
+@property (strong, nonatomic) UIView *incomingCallContainer;
 
 @end
 
 @implementation ActiveCallViewController
 
+@dynamic delegate;
```
dvor: Why do we need `dynamic`?

Antidote/ActiveCallViewController.m
```diff
@@ -89,22 +94,23 @@ - (void)createContainerView
...
 }
 
 - (void)createMuteButton
 {
-    self.muteButton = [self createButtonWithImageName:@"call-audio-enable" action:@selector(toggleMute)];
+    self.speakerButton = [self createButtonWithImageName:@"call-audio-enable" action:@selector(speakerButtonPressed)];
 
     UIImage *selectedImage = [UIImage imageNamed:@"call-audio-disable"];
-    [self.muteButton setImage:selectedImage forState:UIControlStateSelected];
+    [self.speakerButton setImage:selectedImage forState:UIControlStateSelected];
```
dvor: microphoneButton uses `enable` image for selected state, speakerButton uses `disable`. Maybe it would be better to use same approach in both cases?

Antidote/ActiveCallViewController.m
```diff
@@ -153,17 +159,53 @@ - (void)installConstraints
...
+        [self setupIncomingCallView];
+    }
+    else {
+        [self.incomingCallContainer removeFromSuperview];
+        self.incomingCallContainer = nil;
+    }
+
+    _showIncomingCallView = showIncomingCallView;
+}
+
```
dvor: Also there should be method `setIncomingCallCallerName:`

Antidote/ActiveCallViewController.m
```diff
@@ -153,17 +159,53 @@ - (void)installConstraints
...
+- (BOOL)speakerSelected
+{
+    return self.speakerButton.selected;
+}
+
+- (void)setShowIncomingCallView:(BOOL)showIncomingCallView
+{
+    if (showIncomingCallView) {
+        [self setupIncomingCallView];
```
dvor: It would be better to say `createIncomingCallView`.

Also what will happen if `setShowIncomingCallView:YES` method will be called twice?

Antidote/ActiveCallViewController.m
```diff
@@ -186,41 +228,109 @@ - (UIButton *)createButtonWithImageName:(NSString *)imageName action:(SEL)action
...
-    return [NSString stringWithFormat:@"%02d:%02d", minutes, seconds];
+    [nameLabel makeConstraints:^(MASConstraintMaker *make) {
+        make.left.equalTo(self.incomingCallContainer.left);
+        make.height.equalTo(20.0);
+        make.bottom.equalTo(self.incomingCallContainer.centerY);
+    }];
+
+    [descriptionLabel makeConstraints:^(MASConstraintMaker *make) {
+        make.centerX.equalTo(nameLabel);
+        make.left.equalTo(self.incomingCallContainer.left);
```
dvor: I think that `centerX` and `left` constraints are in conflict.

Antidote/DialingCallViewController.m
```diff
@@ -33,20 +29,9 @@ @interface DialingCallViewController ()
...
-#pragma mark - Life cycle
-
-- (instancetype)initWithChat:(OCTChat *)chat submanagerCalls:(OCTSubmanagerCalls *)manager
-{
-    OCTCall *call = [manager callToChat:chat enableAudio:YES enableVideo:NO error:nil];
-
-    if (! call) {
-        return nil;
-    }
+@dynamic delegate;
```
dvor: Do we need this?

Antidote/RingingCallViewController.m
```diff
@@ -39,6 +36,8 @@ @interface RingingCallViewController ()
 
 @implementation RingingCallViewController
 
+@dynamic delegate;
```
dvor: Do we need this?

Antidote/AbstractCallViewController.h
Chuongv: We can remove it but then it would make it less convenient for this method below in Calls Manager
```
- (AbstractCallViewController *)viewControllerForCall:(OCTCall *)call
{
    AbstractCallViewController *viewController;

    OCTFriend *friend = [call.chat.friends firstObject];

    switch (call.status) {
        case OCTCallStatusActive:
            viewController = [[ActiveCallViewController alloc] initWithCallerNickname:friend.nickname];
            break;
        case OCTCallStatusDialing:
            viewController = [[DialingCallViewController alloc] initWithCallerNickname:friend.nickname];
            break;
        case OCTCallStatusRinging:
            viewController = [[RingingCallViewController alloc] initWithCallerNickname:friend.nickname];
            break;
        case OCTCallStatusPaused:
            NSAssert(NO, @"We should not be here yet. Not yet implemented");
            break;
    }

    viewController.modalInPopover = YES;
    viewController.modalPresentationStyle = UIModalPresentationOverCurrentContext;
    viewController.delegate = self;

    return viewController;
}
```

Without the delegate in `Abstract` I would not be able to assign a delegate as easily. Which is why I use `dynamic` in the other subclasses. Perhaps there's a better way?

Antidote/AbstractCallViewController.h
Chuongv: Right now if there's going to be a new Active call, for example when going from Active --> Active, the view controller gets recreated. I guess I could avoid that reusing the same controller and just changing the nickname. I'll try and do that then.

Antidote/ActiveCallViewController.m
Chuongv: You mean a customized setter for `incomingCallCallerName`? What do we need that for though?

Antidote/ActiveCallViewController.m
Chuongv: CallsManager should already be avoiding that situation. However I am not sure if I should also find a way to protect that from happening here?

Antidote/DialingCallViewController.m
Chuongv: If I remove this, it will give me a warning "Auto property synthesis will not synthesize property 'delegate'; it will be implemented by its superclass, use @dynamic to acknowledge intention" Which we do want, since we are assigning the delegate at the superclass `AbstractCallViewController`

Antidote/AbstractCallViewController.h
dvor: It wouldn't be less convenient as compiler won't check if CallsManager confirms to specific protocol. Basically it is better to avoid dynamic behavior when possible, compiler is your friend.

I think that following will be fine
```
switch (call.status) {
    case OCTCallStatusActive: {
        ActiveCallViewController *activeVC = [ActiveCallViewController new];
        activeVC.delegate = self;
        viewController = activeVC;
        break;
    }
    case OCTCallStatusDialing: {
        DialingCallViewController *dialingVC = [DialingCallViewController new];
        dialingVC.delegate = self;
        viewController = dialingVC;
        break;
    }
    case OCTCallStatusRinging: {
        RingingCallViewController *ringingVC = [RingingCallViewController new];
        ringingVC.delegate = self;
        viewController = ringingVC;
        break;
    }
    case OCTCallStatusPaused:
        NSAssert(NO, @"We should not be here yet. Not yet implemented");
        break;
}
```

Antidote/ActiveCallViewController.m
dvor: API allows you set `showIncomingCallView` to `YES`, and then assign `incomingCallCallerName`. However this wouldn't work. I don't think it is good idea to have order-dependent properties.

Antidote/ActiveCallViewController.m
dvor: You cannot know what will happen in a month or in a year. CallsManager can be dead at this point. If you can easily save future yourself (or someone else) from bugs, why not to do that? :-)

Just add check at beginning of method
```
if (showIncomingCallView == _showIncomingCallView) {
    return;
}
```

Antidote/ActiveCallViewController.m
Chuongv: Ah I see, so roll both of them into one method.
`- (void)showIncomingCallViewForFriend:(NSString *)nickname;`

Antidote/CallsManager.m
```diff
@@ -0,0 +1,331 @@
...
+#import "RBQFetchedResultsController.h"
+#import "ProfileManager.h"
+#import "Helper.h"
+#import "OCTCall.h"
+#import "TabBarViewController.h"
+#import "AbstractCallViewController.h"
+
+@interface CallsManager () <RBQFetchedResultsControllerDelegate, ActiveCallViewControllerDelegate, DialingCallViewControllerDelegate, RingingCallViewControllerDelegate>
+
+@property (strong, nonatomic) RBQFetchedResultsController *currentCallController;
```
dvor: `currentCallController` is unused. Afaik you have `allActiveCallsController` instead of it.

Antidote/CallsManager.m
```diff
@@ -0,0 +1,331 @@
...
+@property (strong, nonatomic) RBQFetchedResultsController *allCallsController;
+@property (strong, nonatomic) RBQFetchedResultsController *allActiveCallsController;
+@property (strong, nonatomic) RBQFetchedResultsController *allPausedCallsController;
```
dvor: `allPausedCallsController` is unused

Antidote/ActiveCallViewController.m
dvor: Yes, sounds even better. :+1:

dvor: Also you would need `hide...` method

Antidote/CallsManager.m
```diff
@@ -0,0 +1,331 @@
...
+#import "ActiveCallViewController.h"
+#import "AppDelegate+Utilities.h"
+#import "RBQFetchedResultsController.h"
+#import "ProfileManager.h"
+#import "Helper.h"
+#import "OCTCall.h"
+#import "TabBarViewController.h"
+#import "AbstractCallViewController.h"
+
+@interface CallsManager () <RBQFetchedResultsControllerDelegate, ActiveCallViewControllerDelegate, DialingCallViewControllerDelegate, RingingCallViewControllerDelegate>
```
dvor: It would be nice to have some detailed logging for CallsManager.

It is not part of this pull request, just to let you know for the future.

Antidote/CallsManager.m
```diff
@@ -0,0 +1,331 @@
...
+{
+    if ([self.currentCallViewController isKindOfClass:[ActiveCallViewController class]]) {
+        ActiveCallViewController *activeVC = (ActiveCallViewController *)self.currentCallViewController;
+        activeVC.callDuration = duration;
+    }
+}
+
+- (void)setEnableMicrophone:(BOOL)enableMicrophone
+{
+    self.manager.enableMicrophone = enableMicrophone;
```
dvor: Here we should update UI

Antidote/CallsManager.m
```diff
@@ -0,0 +1,331 @@
...
+    controller.incomingCallCallerName = nil;
+    controller.createIncomingCallView = NO;
+}
+
+- (void)activeCallAnswerIncomingCallButtonPressed:(ActiveCallViewController *)controller
+{
+    [self.manager answerCall:self.pendingIncomingCall enableAudio:YES enableVideo:NO error:nil];
+    self.pendingIncomingCall = nil;
+    controller.incomingCallCallerName = nil;
+    controller.createIncomingCallView = NO;
```
dvor: objcTox will pause active call, right?

Antidote/CallsManager.m
```diff
@@ -0,0 +1,331 @@
...
+- (BOOL)enableMicrophone
+{
+    return self.manager.enableMicrophone;
+}
+
+#pragma mark - ActiveCallViewController Delegate
+
+- (void)activeCallDeclineButtonPressed:(ActiveCallViewController *)controller
+{
+    [self.manager sendCallControl:OCTToxAVCallControlCancel toCall:self.currentCall error:nil];
```
dvor: If you had paused calls you should switch to one of them

Antidote/CallsManager.m
```diff
@@ -0,0 +1,331 @@
...
+    [self.manager answerCall:self.currentCall
+                 enableAudio:YES
+                 enableVideo:NO error:nil];
+}
+
+- (void)ringingCallDeclineButtonPressed:(RingingCallViewController *)controller
+{
+    [self.manager sendCallControl:OCTToxAVCallControlCancel
+                           toCall:self.currentCall
+                            error:nil];
```
dvor: Hm, I found this API a bit confusing. For canceling call you use `OCTToxAVCallControlCancel`, but for answering you send `answerCall:` method. Maybe you can refactor it in objcTox later.

Antidote/CallsManager.m
Chuongv: I think I will use that for when I implement the array of paused calls?

Antidote/CallsManager.m
```diff
@@ -0,0 +1,331 @@
...
+
+    ActiveCallViewController *activeVC = (ActiveCallViewController *)self.currentCallViewController;
+    activeVC.incomingCallCallerName = friend.nickname;
+    activeVC.createIncomingCallView = YES;
+}
+
+- (void)dealloc
+{
+    [self.currentCallViewController dismissViewControllerAnimated:YES completion:nil];
+}
```
dvor: Can you please move dealloc right after init method?

Antidote/CallsManager.m
dvor: Yes, I see it now.

Antidote/CallsManager.m
Chuongv: You think it's best to switch and resume? Currently we have no option to resume a pause call from the UI itself. Or should we add an extra button for pausing/resuming?

Antidote/CallsManager.m
Chuongv: So if you try and answer a call while already in a call, it will automatically pause the current call.

Antidote/CallsManager.m
```diff
@@ -0,0 +1,331 @@
...
+    return viewController;
+}
+
+- (void)dismissAndSwitchViewControllerForCall:(OCTCall *)call
+{
+    AbstractCallViewController *abstractVC = [self viewControllerForCall:call];
+
+    TabBarViewController *tabBarVC = (TabBarViewController *)[AppContext sharedContext].tabBarController;
+
+    if ([tabBarVC presentedViewController]) {
```
dvor: Hm... I'm not happy with that. There can be some other controller presented, not necessary calls. I think itwould be better to check if there is `currentCallViewController` and dismiss it.

Or you can use NavigationController with one controller is stack. When you need to change controllers just replace `viewControllers` property with new controller

Antidote/CallsManager.m
Chuongv: Yea it is a bit.  The reason is because in ToxAV they are two separate methods. There is no `sendCallControlAnswer` for answering calls. Which would of been nice.

Antidote/CallsManager.m
dvor: I think that play/pause button would be fine. In case if active call has ended you can show paused call with pause button pressed.

Antidote/CallsManager.m
dvor: Good

Antidote/CallsManager.m
Chuongv: I like the navigationController idea.

Antidote/CallsManager.m
Chuongv: Yea good idea

Chuongv:
Not sure if I am quite happy on the implementation of the PausedCallsTableView. There's probably a better way of doing this?

Right now I have `CallsManager` providing the `tableView` to `ActiveCallViewController`.

Also let me know if you would like for me to move the start of this implementation to a different PR. Since this one is getting a little big.

dvor:
> Also let me know if you would like for me to move the start of this implementation to a different PR. Since this one is getting a little big.

Yeah, don't push new code here. Lets leave this pull request only for code review fixes.

Antidote/AbstractCallViewController.m
```diff
@@ -29,85 +30,91 @@ @implementation AbstractCallViewController
...
-- (void)controllerDidChangeContent:(RBQFetchedResultsController *)controller
+- (void)createNameLabel
 {
-    self.call = [[self.callController fetchedObjects] firstObject];
+    self.nameLabel = [UILabel new];
+    self.nameLabel.text = self.nickname;
+    self.nameLabel.font = [[AppContext sharedContext].appearance fontHelveticaNeueWithSize:kNameLabelFontSize];
+    self.nameLabel.textColor = [UIColor whiteColor];
+    self.nameLabel.textAlignment = NSTextAlignmentCenter;
+    [self.nameLabel sizeToFit];
```
dvor: I don't think that you need to call `sizeToFit`. It should resize automatically with autolayout.

Antidote/AbstractCallViewController.m
```diff
@@ -29,85 +30,91 @@ @implementation AbstractCallViewController
...
-    if (! self.call) {
-        [self.navigationController dismissViewControllerAnimated:YES completion:nil];
-    }
+- (void)createSublabel
+{
+    self.subLabel = [UILabel new];
+    self.subLabel.font = [[AppContext sharedContext].appearance fontHelveticaNeueWithSize:kSublabelFontSize];
+    self.subLabel.textColor = [UIColor whiteColor];
+    self.subLabel.textAlignment = NSTextAlignmentCenter;
+    [self.subLabel sizeToFit];
```
dvor: Same about this `sizeToFit`

Antidote/CallsManager.m
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    viewController.nickname = friend.nickname;
+    viewController.modalInPopover = YES;
+    viewController.modalPresentationStyle = UIModalPresentationOverCurrentContext;
+
+    return viewController;
+}
+
+- (void)dismissAndSwitchViewControllerForCall:(OCTCall *)call
```
dvor: This method isn't dismissing anything now. I think with should be called `switchViewControllerForCall:`

Antidote/CallsManager.m
```diff
@@ -0,0 +1,401 @@
...
+    AbstractCallViewController *abstractVC = [self viewControllerForCall:call];
+
+    [self.callNavigation setViewControllers:@[abstractVC] animated:YES];
+
+    self.currentCallViewController = abstractVC;
+
+    self.currentCall = call;
+}
+
+- (void)dismissAndSwitchToActiveViewControllerForCall:(OCTCall *)call
```
dvor: I'm not sure if we need separate method for that. You can move this logic to `switchViewControllerForCall:`.

This way you can simple call `switch...` and doesn't care if you switched to active or non-active view controller

Antidote/CallsManager.m
```diff
@@ -0,0 +1,401 @@
...
+{
+    if (self.currentCall) {
+        NSAssert(NO, @"We should not be able to make another call yet");
+        return;
+    }
+
+    OCTCall *call = [self.manager callToChat:chat enableAudio:YES enableVideo:NO error:nil];
+
+    if (! call) {
+        [[AppContext sharedContext] killCallsManager];
```
dvor: Probably return here?

Antidote/CallsManager.m
```diff
@@ -0,0 +1,401 @@
...
+
+#pragma mark - RBQFetchedResultsControllerDelegate
+
+- (void) controller:(RBQFetchedResultsController *)controller
+    didChangeObject:(RBQSafeRealmObject *)anObject
+        atIndexPath:(NSIndexPath *)indexPath
+      forChangeType:(NSFetchedResultsChangeType)type
+       newIndexPath:(NSIndexPath *)newIndexPath;
+{
+    if (type == NSFetchedResultsChangeUpdate) {
```
dvor: It would be nicer to change those `if`'s to `switch`

```
switch (type) {
   case NSFetchedResultsChangeUpdate: {
        // code
        break;
    }
    case NSFetchedResultsChangeDelete: {
...
```

Antidote/CallsManager.m
```diff
@@ -0,0 +1,401 @@
...
+        if (controller == self.allActiveCallsController) {
+            OCTCall *call = [anObject RLMObject];
+            [self updateDuration:call.callDuration];
+        }
+    }
+
+    if (type == NSFetchedResultsChangeDelete) {
+        if ([self.allCallsController.fetchedObjects count] == 0) {
+            [[AppContext sharedContext] killCallsManager];
```
dvor: `return;` after killing manager

Antidote/CallsManager.m
dvor: This is still relevant.

dvor:
> Not sure if I am quite happy on the implementation of the PausedCallsTableView. There's probably a better way of doing this?

Yes, there is. First of all I propose you to completely remove `PausedCallsTableView` from this pull request (otherwise we'll never merge it ;-) ).

In another pull request you can
- create ActiveCallVC dataSource. It will have following methods
  - `- (NSUInteger)abstractControllerNumberOfPausedCalls:...`
  - `- (NSString *)abstractController:... pausedCallerNicknameForCallAtIndex:(NSUInteger)index;`
  - `- (NSDate *)abstractController:... pausedStartDateForCallAtIndex:(NSUInteger)index;`
- controller will use these methods to populate tableView
- controller will use number of paused calls to show/hide top right button
- controller can have method like `reloadPausedCalls`
- controller's delegate will have `abstractController:... didSelectPausedCallAtIndex:` to notify CallsManager

Chuongv:
Should the protocol be in `AbstractCalVC` or `ActiveCallVC`? You mention `ActiveCallVC` but in the protocol methods they are `abstractController`.

Antidote/CallsManager.m
Chuongv: Oops missed this one. Looks like this method is actually unused. Removing it now.

Chuongv: It's actually taken care of in 
`- (void)activeCallMicButtonPressed:(ActiveCallViewController *)controller`

dvor:
Hm, paused calls will be accessible only from ActiveCallVC (at least for now). So you can put protocol in active view controller

PR #44, opened 2015-07-17
Add Pause Menu Table View

Chuongv:
I think I got most of the pause call menu functionality down. Only thing missing is that it needs to dismiss a view if a touch is triggered outside the tableview.

Had some trouble positioning the red badge on top of pause menu button. So if you know a better way to improve it, let me know :smile:


Antidote/ActiveCallViewController.h
```diff
@@ -23,7 +23,16 @@
...
+
+@end
+
+@protocol ActiveCallViewControllerDataSource <NSObject>
+
+- (NSInteger)activeCallControllerNumberOfPausedCalls:(ActiveCallViewController *)controller;
+- (NSString *)activeCallController:(ActiveCallViewController *)controller pausedCallerNicknameForCallAtIndex:(NSIndexPath *)indexPath;
+- (NSTimeInterval)activeCallController:(ActiveCallViewController *)controller pauseTimeDurationForCallAtIndex:(NSIndexPath *)indexPath;
+- (void)activeCallController:(ActiveCallViewController *)controller resumePausedCallSelectedAtIndex:(NSIndexPath *)indexPath;
+- (void)activeCallController:(ActiveCallViewController *)controller endPausedCallSelectedAtIndex:(NSIndexPath *)indexPath;
```
dvor: Hm, there is no need in indexPath in these methods, as we don't use sections. You can simply pass `NSUInteger` with index.

Antidote/ActiveCallViewController.m
```diff
@@ -156,6 +209,30 @@ - (void)installConstraints
...
+        make.width.equalTo(kButtonSide);
+        make.height.equalTo(kButtonSide);
+        make.right.equalTo(self.topViewContainer.rightMargin);
+        make.top.equalTo(self.topViewContainer.topMargin);
+    }];
+
+    [self.tableViewOfPausedCalls makeConstraints:^(MASConstraintMaker *make) {
+        make.top.equalTo(self.topViewContainer.bottom);
+        make.width.equalTo(self.view);
+        make.height.equalTo(self.view.bounds.size.height / 3.0);
```
dvor: I'm not sure it this will work after you rotate device. I think it is better instead to set `bottom` constraint to `(self.view.bottom).offset(-kSomeConstant);`

Antidote/ActiveCallViewController.m
```diff
@@ -156,6 +209,30 @@ - (void)installConstraints
...
+    }];
+
+    [self.badgeContainer makeConstraints:^(MASConstraintMaker *make) {
+        make.top.equalTo(self.callMenuButton.centerY);
+        make.centerX.equalTo(self.callMenuButton.imageView.frame.origin.x);
```
dvor: Instead of `centerX`
```
make.right.equalTo(self.callMenuButton.right).offset(-kSomeNiceConstant);
```

Antidote/ActiveCallViewController.m
```diff
@@ -211,6 +288,13 @@ - (void)hideIncomingCallView
     self.incomingCallContainer = nil;
 }
 
+- (void)reloadPausedCalls
+{
+    self.badgeLabel.text = [NSString stringWithFormat:@"%lu", [self.dataSource activeCallControllerNumberOfPausedCalls:self]];
```
dvor: I think we should hide badge button if there is no paused calls.

Antidote/ActiveCallViewController.m
```diff
@@ -310,6 +394,64 @@ - (void)setupIncomingCallViewForFriend:(NSString *)nickname
...
+#pragma mark - Call Menu
+
+- (void)callMenuButtonPressed
+{
+    if (self.tableViewOfPausedCalls.hidden) {
+        self.tableViewOfPausedCalls.hidden = NO;
+    }
+    else {
+        self.tableViewOfPausedCalls.hidden = YES;
+    }
```
dvor:
```
self.tableViewOfPausedCalls.hidden = ! self.tableViewOfPausedCalls.hidden;
```

Antidote/ActiveCallViewController.m
```diff
@@ -310,6 +394,64 @@ - (void)setupIncomingCallViewForFriend:(NSString *)nickname
...
+}
+
+#pragma mark - UITableViewDataSource
+
+- (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    PauseCallTableViewCell *cell = [self.tableViewOfPausedCalls dequeueReusableCellWithIdentifier:[PauseCallTableViewCell reuseIdentifier]];
+
+    NSString *nickName = [self.dataSource activeCallController:self
+                            pausedCallerNicknameForCallAtIndex             :indexPath];
```
dvor: Uncrustify goes wild sometimes :\

Antidote/ActiveCallViewController.m
```diff
@@ -310,6 +394,64 @@ - (void)setupIncomingCallViewForFriend:(NSString *)nickname
...
+    cell.delegate = self;
+
+    return cell;
+}
+
+- (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger)section
+{
+    NSInteger numberOfPausedCalls = [self.dataSource activeCallControllerNumberOfPausedCalls:self];
+
+    self.callMenuButton.hidden = (numberOfPausedCalls == 0);
```
dvor: Why you update button here? You should move this method to `reloadPausedCalls`.

Also you need to hide badge.

dvor: Oh, I see, badge is added to button as subview. I think it is better to just put it on top. This way you can move it to the right, out of button's frame

Antidote/ActiveCallViewController.m
```diff
@@ -310,6 +394,64 @@ - (void)setupIncomingCallViewForFriend:(NSString *)nickname
...
+
+#pragma mark PauseCallTableViewCellDelegate
+
+- (void)pauseCallCellEndPausedCallButtonTapped:(PauseCallTableViewCell *)cell
+{
+    NSIndexPath *indexPath = [self.tableViewOfPausedCalls indexPathForCell:cell];
+
+    [self.dataSource activeCallController:self endPausedCallSelectedAtIndex:indexPath];
+
+    self.tableViewOfPausedCalls.hidden = YES;
```
dvor: It would be better for user to hide paused calls only if there are no paused calls left. That way user will be able to dismiss several calls at one time.

dvor: You can simply hide (but not show!) tableView in `reloadPausedCalls` method (if there are no more paused calls available). This way you'll also handle case when last paused partner canceled the call

Antidote/CallsManager.m
```diff
@@ -20,7 +20,7 @@
 #import "TabBarViewController.h"
 #import "AbstractCallViewController.h"
 
-@interface CallsManager () <RBQFetchedResultsControllerDelegate, ActiveCallViewControllerDelegate, DialingCallViewControllerDelegate, RingingCallViewControllerDelegate>
+@interface CallsManager () <RBQFetchedResultsControllerDelegate, ActiveCallViewControllerDelegate, ActiveCallViewControllerDataSource, DialingCallViewControllerDelegate, RingingCallViewControllerDelegate>
```
dvor: Can you please break delegates to several lines?

Antidote/CallsManager.m
```diff
@@ -271,6 +275,43 @@ - (void)ringingCallDeclineButtonPressed:(RingingCallViewController *)controller
                             error:nil];
 }
 
+#pragma mark - ActiveCallViewControllerDataSource
+
+- (NSInteger)activeCallControllerNumberOfPausedCalls:(ActiveCallViewController *)controller
+{
+    return [[self.allPausedCallsController fetchedObjects] count];
```
dvor: `return [self.allPausedCallsController numberOfRowsForSectionIndex:0];` instead

Antidote/CallsManager.m
```diff
@@ -271,6 +275,43 @@ - (void)ringingCallDeclineButtonPressed:(RingingCallViewController *)controller
...
+}
+
+- (NSString *)activeCallController:(ActiveCallViewController *)controller pausedCallerNicknameForCallAtIndex:(NSIndexPath *)indexPath
+{
+    OCTCall *call = [self.allPausedCallsController objectAtIndexPath:indexPath];
+
+    OCTFriend *friend = [call.chat.friends firstObject];
```
dvor: We have `call.caller` property for that.

dvor: Oh, sorry, if you started the call then `caller` will be nil. Then everything is correct.

Antidote/CallsManager.m
```diff
@@ -271,6 +275,43 @@ - (void)ringingCallDeclineButtonPressed:(RingingCallViewController *)controller
...
+
+    return friend.nickname;
+}
+
+- (NSTimeInterval)activeCallController:(ActiveCallViewController *)controller pauseTimeDurationForCallAtIndex:(NSIndexPath *)indexPath
+{
+    OCTCall *call = [self.allPausedCallsController objectAtIndexPath:indexPath];
+
+    return call.callDuration;
```
dvor: There should be not `callDuration`, but `on hold duration`. So what I suggest to you is to store `on hold start time` and pass it to controller. Controller will have timer that will tick every second, which will be used for updating pause cells.

Antidote/PauseCallTableViewCell.m
```diff
@@ -0,0 +1,123 @@
...
+
+    [self.contentView addSubview:self.endCallButton];
+}
+
+- (void)createResumeCallButton
+{
+    UIImage *image = [UIImage imageNamed:@"call-pause"];
+
+    [self.imageView setImage:image];
+    self.imageView.tintColor = [UIColor whiteColor];
```
dvor: Where is the button? :-)

Antidote/PauseCallTableViewCell.m
```diff
@@ -0,0 +1,123 @@
...
+    [self installConstraints];
+
+    return self;
+}
+
+#pragma mark - View setup
+
+- (void)createEndCallButton
+{
+    UIImage *image = [UIImage imageNamed:@"call-accept"];
```
dvor: Why "endCall" has "accept" image?

Antidote/PauseCallTableViewCell.m
```diff
@@ -0,0 +1,123 @@
...
+        make.top.equalTo(self.contentView);
+        make.bottom.equalTo(self.contentView.centerY);
+    }];
+
+    [self.detailTextLabel makeConstraints:^(MASConstraintMaker *make) {
+        make.left.equalTo(self.imageView.right).with.offset(10.0);
+        make.right.lessThanOrEqualTo(self.endCallButton.left);
+        make.top.equalTo(self.contentView.centerY);
+        make.bottom.equalTo(self.contentView);
+    }];
```
dvor: Have you checked if your constraints conflict with standard ones? Have you checked if it is working correctly for long names?

Chuongv: What do you mean conflicts with standard ones? I haven't checked with long names yet, that's a good idea. I'll need to do that.

Antidote/PauseCallTableViewCell.m
Chuongv: The image was originally supposed to be used for accepting calls, but looks like it's only used for declining calls. I'll need to rename and change it throughout the project.

Antidote/ActiveCallViewController.h
Chuongv: Not sure what you mean? The indexPath is needed because the `RBQFetchedResultsController` has a method for `objectAtIndexPath:indexPath` which I use in those methods.

`OCTCall *call = [self.allPausedCallsController objectAtIndexPath:indexPath];`

Antidote/PauseCallTableViewCell.m
dvor: I mean that there are default constraints for both textLabel and detailedTextLabel. You just have to check if everything if working fine. There can be three kind of problems:
- wrong layout
- warning message about broken constraints in console
- crash :-)

Antidote/ActiveCallViewController.h
dvor: Ah, okay then.

Antidote/CallsManager.m
Chuongv: Do you think this would be better done at the objcTox level instead?

Antidote/PauseCallTableViewCell.m
Chuongv: Yea haha. Realized halfway through that I didn't need a button when I could just use the imageView. Will change it.

Antidote/CallsManager.m
dvor: `onHoldStartDate` can be done in objcTox, yes.

Chuongv: I was thinking of a timer :smile: . Though `onHoldStartDate` could work, though I'm not sure what would be the best way to refresh the tableView. Also I thought saving realm objects, you lose precision on time with NSDate.

dvor: I don't think that timer should go to objcTox. So as I see it:
- add `onHoldStartDate` to objcTox (or `onHoldStartInterval` for saving precision)
- in ActiveCallVC, when table is visible start timer, that will tick every second
- when table isn't visible, remove timer.

Chuongv: Ah good idea. :+1:

PR #45, opened 2015-07-17
Add logging to CallsManager

Chuongv:
Logging for CallsManager

### DIFF
--- a/Antidote/CallsManager.m
+++ b/Antidote/CallsManager.m
@@ -214,7 +214,7 @@
 
     NSError *error;
 
-    OCTToxAVCallControl control = (controller.speakerSelected)? OCTToxAVCallControlUnmuteAudio : OCTToxAVCallControlMuteAudio;
+    OCTToxAVCallControl control = (controller.speakerSelected) ? OCTToxAVCallControlUnmuteAudio : OCTToxAVCallControlMuteAudio;
 
     if ([self.manager sendCallControl:control toCall:self.currentCall error:&error]) {
         controller.speakerSelected = ! controller.speakerSelected;
@@ -352,7 +352,7 @@
     AALogVerbose(@"%@ call:%@", controller, call);
 
     NSError *error;
-    if (! [self.manager sendCallControl:OCTToxAVCallControlCancel toCall:call error:&error] ) {
+    if (! [self.manager sendCallControl:OCTToxAVCallControlCancel toCall:call error:&error]) {
         AALogWarn(@"%@", error);
     }
 }
--- a/Antidote/CallsManager.m
+++ b/Antidote/CallsManager.m
@@ -20,7 +20,7 @@
 #import "TabBarViewController.h"
 #import "AbstractCallViewController.h"
 
-#define LOG_IDENTIFIER @"CallsManager"
+#define LOG_IDENTIFIER self
 
 @interface CallsManager () <RBQFetchedResultsControllerDelegate, ActiveCallViewControllerDelegate,
                             ActiveCallViewControllerDataSource, DialingCallViewControllerDelegate,
@@ -45,7 +45,7 @@
 #pragma mark - Lifecycle
 - (instancetype)init
 {
-    AALogVerbose(@"");
+    AALogVerbose();
     self = [super self];
 
     if (! self) {
@@ -80,7 +80,7 @@
 
 - (void)dealloc
 {
-    AALogVerbose(@"");
+    AALogVerbose();
     [self.callNavigation dismissViewControllerAnimated:YES completion:nil];
 }
 
@@ -99,7 +99,7 @@
     OCTCall *call = [self.manager callToChat:chat enableAudio:YES enableVideo:NO error:&error];
 
     if (! call) {
-        AALogWarn(@"error %@, failureReason:%@", error, error.localizedFailureReason);
+        AALogWarn(@"%@", error);
         [[AppContext sharedContext] killCallsManager];
         return;
     }
@@ -195,7 +195,7 @@
 
     NSError *error;
     if (! [self.manager sendCallControl:OCTToxAVCallControlCancel toCall:self.currentCall error:&error]) {
-        DDLogWarn(@"%@, activeCallDeclineButtonPressed error: %@", self, error.localizedFailureReason);
+        AALogWarn(@"%@", error);
     }
 }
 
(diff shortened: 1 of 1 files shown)


Antidote/CallsManager.m
```diff
@@ -75,6 +75,7 @@ - (instancetype)init
 
 - (void)dealloc
```
dvor: Log in init will be useful too. As CallsManager can be killed and recreated it would be useful to track those events.

Antidote/CallsManager.m
```diff
@@ -82,14 +83,18 @@ - (void)dealloc
...
         NSAssert(NO, @"We should not be able to make another call yet");
         return;
     }
 
-    OCTCall *call = [self.manager callToChat:chat enableAudio:YES enableVideo:NO error:nil];
+    NSError *error;
+    OCTCall *call = [self.manager callToChat:chat enableAudio:YES enableVideo:NO error:&error];
 
     if (! call) {
+        DDLogWarn(@"%@, Error callToChat: %@", self, error.localizedFailureReason);
```
dvor: I think that logging `error` would give more information than logging just `error.localizedFailureReason`. Can you change it to `error` in all logs or you have any objections?

Antidote/CallsManager.m
```diff
@@ -205,28 +214,46 @@ - (void)activeCallSpeakerButtonPressed:(ActiveCallViewController *)controller
...
-        if ([self.manager sendCallControl:OCTToxAVCallControlResume toCall:self.currentCall error:nil]) {
+        if ([self.manager sendCallControl:OCTToxAVCallControlResume toCall:self.currentCall error:&error]) {
             controller.pauseSelected = NO;
         }
+        else {
+            DDLogWarn(@"%@, activeCallPauseButtonPressed:OCTToxAVCallControlResume error: %@", self, error.localizedFailureReason);
+        }
     }
     else {
         if ([self.manager sendCallControl:OCTToxAVCallControlPause toCall:self.currentCall error:nil]) {
```
dvor: Here you missed `&error` parameter. To avoid such mistakes it's better to avoid code duplication. :-) You can refactor this method like this:

```
- (void)activeCallPauseButtonPressed:(ActiveCallViewController *)controller
{
    NSError *error;
    OCTToxAVCallControl control = controller.pauseSelected ? OCTToxAVCallControlResume : OCTToxAVCallControlPause;

    BOOL result = [self.manager sendCallControl:control toCall:self.currentCall error:&error];

    if (result) {
        controller.pauseSelected = ! controller.pauseSelected;
    }
    else {
        DDLogWarn(@"%@, activeCallPauseButtonPressed:OCTToxAVCallControlPause error: %@", self, error);
    }
}
```

Antidote/CallsManager.m
```diff
@@ -205,28 +214,46 @@ - (void)activeCallSpeakerButtonPressed:(ActiveCallViewController *)controller
...
+        else {
+            DDLogWarn(@"%@, activeCallPauseButtonPressed:OCTToxAVCallControlPause error: %@", self, error.localizedFailureReason);
+        }
     }
 }
 
 - (void)activeCallDeclineIncomingCallButtonPressed:(ActiveCallViewController *)controller
 {
-    [self.manager sendCallControl:OCTToxAVCallControlCancel toCall:self.pendingIncomingCall error:nil];
+    NSError *error;
```
dvor: Here you can add verbose log as for `activeCallAnswerIncomingCallButtonPressed` method.

dvor:
Please add logs to `- handleIncomingCall:` method, it is quite important.

I've added macros for logging https://github.com/Antidote-for-Tox/Antidote/commit/6c26a4e7cd905d375eb2214cc7bde6dcb088e68f. Please replace all you logs with them.

Now you don't have to pass every time `self` as parameter nor specify method name. Instead you have to define `LOG_IDENTIFIER` at the beginning of file
```
#defile LOG_IDENTIFIER self
```
and then just replace this
```
DDLogVerbose(@"%@, switchViewControllerForCall:%@", self, call);
DDLogWarn(@"%@, sendCallControl:OCTToxAVCallControlCancel error:%@", self, error.localizedFailureReason);
```
with this
```
AALogVerbose(@"%@", call);
AALogWarn(@"OCTToxAVCallControlCancel error:%@", error);
```

Antidote/CallsManager.m
Chuongv: Ah this is so much cleaner. Thanks :smile:

Antidote/CallsManager.m
```diff
@@ -20,6 +20,8 @@
 #import "TabBarViewController.h"
 #import "AbstractCallViewController.h"
 
+#define LOG_IDENTIFIER @"CallsManager"
```
dvor: It would be better to define identifier as `self` as this way address of CallsManager will be logged. It may be useful to distinguish different instances of manager in logs.

Antidote/CallsManager.m
```diff
@@ -43,6 +45,7 @@ @implementation CallsManager
 #pragma mark - Lifecycle
 - (instancetype)init
 {
+    AALogVerbose(@"");
```
dvor: You can write `AALogVerbose();` which looks nicer :-)

Antidote/CallsManager.m
```diff
@@ -84,14 +88,18 @@ - (void)dealloc
...
         NSAssert(NO, @"We should not be able to make another call yet");
         return;
     }
 
-    OCTCall *call = [self.manager callToChat:chat enableAudio:YES enableVideo:NO error:nil];
+    NSError *error;
+    OCTCall *call = [self.manager callToChat:chat enableAudio:YES enableVideo:NO error:&error];
 
     if (! call) {
+        AALogWarn(@"error %@, failureReason:%@", error, error.localizedFailureReason);
```
dvor: Isn't failure reason included in error description, that you pass as first argument? I think it should be there as well as domain and error code.

Antidote/CallsManager.m
```diff
@@ -209,6 +228,7 @@ - (void)activeCallSpeakerButtonPressed:(ActiveCallViewController *)controller
             controller.speakerSelected = YES;
         }
         else if (error.code == OCTToxAVErrorControlInvaldTransition) {
+            DDLogWarn(@"%@, sendCallControl error:%@", self, error.localizedFailureReason);
```
dvor: If code isn't equal to InvalidTransition there will be no log. It would be better to put this log after `else if` block.

Antidote/CallsManager.m
```diff
@@ -292,16 +344,27 @@ - (NSTimeInterval)activeCallController:(ActiveCallViewController *)controller pa
 
 - (void)activeCallController:(ActiveCallViewController *)controller resumePausedCallSelectedAtIndex:(NSIndexPath *)indexPath
 {
+    AALogVerbose(@"%@ indexPath:%@", controller, indexPath);
```
dvor: I think it would be better to log call instead of indexPath

Antidote/CallsManager.m
```diff
@@ -292,16 +344,27 @@ - (NSTimeInterval)activeCallController:(ActiveCallViewController *)controller pa
...
+    NSError *error;
+    if (! [self.manager sendCallControl:OCTToxAVCallControlResume toCall:call error:&error]) {
+        AALogWarn(@"Error: %@, FailureReason: %@, Call:%@", error, error.localizedFailureReason, call);
+    }
+
 }
 
 - (void)activeCallController:(ActiveCallViewController *)controller endPausedCallSelectedAtIndex:(NSIndexPath *)indexPath
 {
+    AALogVerbose(@"%@ indexPath:%@", controller, indexPath);
```
dvor: Same for this indexPath

Antidote/CallsManager.m
Chuongv: There is a lot wrong with this method, I'll refactor it.

Antidote/CallsManager.m
Chuongv: Just tested, you are right. I didn't know that :)

dvor:
uncrustify

PR #46, opened 2015-07-17
Add TableViewRefresher

Chuongv:
I had the `reloadData` done on the main queue otherwise the UI would not update.


dvor:
Yeah, general rule - all stuff with UIKit should be done on main thread.

dvor:
I think instead of TableViewRefresher it would be better to use NSTimer. The reason why we didn't use it in objcTox - it isn't that easy to get it working *not* on main thread. But in our case it would be enough to have property with NSTimer in controller.

Chuongv:
Ah good idea. I'll try NSTimer then.

PR #47, opened 2015-07-18
Call activity shown on allchatsviewcontroller

Chuongv:
This now shows the time stamp on the `AllChatsViewController` like so:

screen shot 2015-07-17 at 6 16 44 pm


dvor:
Cool

PR #49, opened 2015-07-19
Add NSTimer for refreshing pause calls when tableview is shown

Chuongv:
Responsible for reloading the tableViewOfPausedCalls.

### DIFF
--- a/Antidote/ActiveCallViewController.m
+++ b/Antidote/ActiveCallViewController.m
@@ -305,8 +305,7 @@ static const CGFloat kBadgeFontSize = 14.0;
     self.callMenuButton.hidden = (numberOfPausedCalls == 0);
 
     if (numberOfPausedCalls == 0) {
-        self.tableViewOfPausedCalls.hidden = YES;
-        [self.tableViewRefreshTimer invalidate];
+        [self hideTableViewOfPausedCalls];
     }
 
     [self.tableViewOfPausedCalls reloadData];
@@ -411,8 +410,13 @@ static const CGFloat kBadgeFontSize = 14.0;
     }];
 }
 
-- (void)startTableViewRefreshTimer
+- (void)showTableViewOfPausedCalls
 {
+    if (! self.tableViewOfPausedCalls.hidden) {
+        return;
+    }
+
+    self.tableViewOfPausedCalls.hidden = NO;
     self.tableViewRefreshTimer = [NSTimer scheduledTimerWithTimeInterval:1.0
                                                                   target:self
                                                                 selector:@selector(reloadPausedCalls)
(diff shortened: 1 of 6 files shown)


Antidote/ActiveCallViewController.m
```diff
@@ -409,11 +411,27 @@ - (void)setupIncomingCallViewForFriend:(NSString *)nickname
...
 - (void)callMenuButtonPressed
 {
     self.tableViewOfPausedCalls.hidden = ! self.tableViewOfPausedCalls.hidden;
+
+    if (! self.tableViewOfPausedCalls.hidden) {
+        [self startTableViewRefreshTimer];
+    }
+    else {
+        [self.tableViewRefreshTimer invalidate];
+    }
```
dvor: It is better to avoid negation, this makes code more readable.
```
if (self.tableViewOfPausedCalls.hidden) {
    [self.tableViewRefreshTimer invalidate];
}
else {
    [self startTableViewRefreshTimer];
}
```

Antidote/ActiveCallViewController.m
```diff
@@ -422,6 +440,7 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
 {
     [self.dataSource activeCallController:self resumePausedCallSelectedAtIndex:indexPath];
     self.tableViewOfPausedCalls.hidden = YES;
+    [self.tableViewRefreshTimer invalidate];
```
dvor: Even better way to improve this code - create separate methods for showing/hiding tableView in which you also can start/stop timer. This way instead of setting `tableViewOfPausedCalls` to hidden and then invalidating the timer, you'll just call `hide...` method.

```
- (void)showTableViewOfPausedCalls
{
    if (! self.tableViewOfPausedCalls.hidden) {
        return;
    }

    self.tableViewOfPausedCalls.hidden = NO;
    self.tableViewRefreshTimer = [NSTimer scheduledTimerWithTimeInterval:1.0
                                                                  target:self
                                                                selector:@selector(reloadPausedCalls)
                                                                userInfo:nil
                                                                 repeats:YES];
}

- (void)hideTableViewOfPausedCalls
{
    self.tableViewOfPausedCalls.hidden = YES;

    [self.tableViewRefreshTimer invalidate];
    self.tableViewRefreshTimer = nil;
}
```

Here you can replace two lines with
```
[self hideTableViewOfPausedCalls];
```

In `callMenuButtonPressed` method:

```
- (void)callMenuButtonPressed
{
    BOOL wasHidden = self.tableViewOfPausedCalls.hidden;

    wasHidden ? [self showTableViewOfPausedCalls] : [self hideTableViewOfPausedCalls];
}
```

Antidote/ActiveCallViewController.m
```diff
@@ -305,6 +306,7 @@ - (void)reloadPausedCalls
 
     if (numberOfPausedCalls == 0) {
         self.tableViewOfPausedCalls.hidden = YES;
+        [self.tableViewRefreshTimer invalidate];
```
dvor: Here you can also call 
```
[self hideTableViewOfPausedCalls];
return;
```

dvor:
Hm, I think it is failing because of objcTox update

Chuongv:
Hopefully it should work now

PR #50, opened 2015-07-21
Add dismissal of tableview when tapping outside

Chuongv:
Now dismisses when tapping below the table view. Only problem is that it won't forward the touch over to the hangup button if you are tapping on that. Is there a better way to implement this?

### DIFF
--- a/Antidote/ActiveCallViewController.m
+++ b/Antidote/ActiveCallViewController.m
@@ -21,7 +21,7 @@ static const CGFloat k3ButtonGap = 15.0;
 static const CGFloat kOtherCallButtonSize = 4.0 / 5.0 * kButtonSide;
 static const CGFloat kIncomingNameFontSize = 12.0;
 static const CGFloat kIncomingIsCallingFontSize = 10.0;
-static const CGFloat kTableViewBottomOffSet = 100.0;
+static const CGFloat kTableViewBottomOffSet = 200.0;
 
 static const CGFloat kBadgeContainerHorizontalOffset = 10.0;
 static const CGFloat kBadgeHeightWidth = 30.0;
@@ -40,6 +40,8 @@ static const CGFloat kBadgeFontSize = 14.0;
 @property (strong, nonatomic) UIButton *callMenuButton;
 @property (strong, nonatomic) UIView *badgeContainer;
 @property (strong, nonatomic) UILabel *badgeLabel;
+
+@property (strong, nonatomic) UIButton *pauseCallsContainer;
 @property (strong, nonatomic) UITableView *tableViewOfPausedCalls;
 
 @property (strong, nonatomic) NSTimer *tableViewRefreshTimer;
@@ -141,14 +143,23 @@ static const CGFloat kBadgeFontSize = 14.0;
 
 - (void)createCallPauseTableView
 {
+
+    self.pauseCallsContainer = [UIButton new];
+    self.pauseCallsContainer.backgroundColor = [UIColor clearColor];
+    [self.pauseCallsContainer addTarget:self
+                                 action:@selector(hideTableViewOfPausedCalls)
+                       forControlEvents:UIControlEventTouchUpInside];
+    self.pauseCallsContainer.hidden = YES;
+
+    [self.view addSubview:self.pauseCallsContainer];
+
     self.tableViewOfPausedCalls = [[UITableView alloc] initWithFrame:CGRectZero style:UITableViewStylePlain];
     self.tableViewOfPausedCalls.backgroundColor = [UIColor blackColor];
     self.tableViewOfPausedCalls.delegate = self;
     self.tableViewOfPausedCalls.dataSource = self;
-    self.tableViewOfPausedCalls.hidden = YES;
     [self.tableViewOfPausedCalls registerClass:[PauseCallTableViewCell class] forCellReuseIdentifier:[PauseCallTableViewCell reuseIdentifier]];
 
-    [self.view addSubview:self.tableViewOfPausedCalls];
+    [self.pauseCallsContainer addSubview:self.tableViewOfPausedCalls];
 }
 
 - (void)createBadgeViews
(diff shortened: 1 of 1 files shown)


Antidote/ActiveCallViewController.m
```diff
@@ -446,7 +465,7 @@ - (void)callMenuButtonPressed
 - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath *)indexPath
 {
     [self.dataSource activeCallController:self resumePausedCallSelectedAtIndex:indexPath];
-    self.tableViewOfPausedCalls.hidden = YES;
+    self.pauseCallsContainer.hidden = YES;
     [self.tableViewRefreshTimer invalidate];
```
dvor: Hm, here you should call `hideTableViewOfPausedCalls` instead of hiding and invalidating timer.

dvor:
> Only problem is that it won't forward the touch over to the hangup button if you are tapping on that. Is there a better way to implement this?

I think that is just fine. If user taps on hangup button, the table hides. Then user can tap on hangup button once again and end call.

dvor:
Or we are talking about hangup button in tableView?

Anyway I don't think it is good to add tableView as a subview to button. I think it would be better to
- create transparent container
- set it `edges` to edges of a `view`
- add button there
- on top of button add tableView

Antidote/ActiveCallViewController.m
```diff
@@ -412,10 +429,11 @@ - (void)setupIncomingCallViewForFriend:(NSString *)nickname
 
 - (void)showTableViewOfPausedCalls
 {
-    if (! self.tableViewOfPausedCalls.hidden) {
+    if (! self.pauseCallsContainer.hidden) {
         return;
     }
 
+    self.pauseCallsContainer.hidden = NO;
     self.tableViewOfPausedCalls.hidden = NO;
```
dvor: No need to hide/show tableView, container would be enough.

Antidote/ActiveCallViewController.m
```diff
@@ -426,6 +444,7 @@ - (void)showTableViewOfPausedCalls
 
 - (void)hideTableViewOfPausedCalls
 {
+    self.pauseCallsContainer.hidden = YES;
     self.tableViewOfPausedCalls.hidden = YES;
```
dvor: No need to hide/show tableView, container would be enough.

dvor:
> This branch has conflicts that must be resolved

Antidote/ActiveCallViewController.m
```diff
@@ -141,15 +144,28 @@ - (void)createCallMenuButton
...
     self.tableViewOfPausedCalls = [[UITableView alloc] initWithFrame:CGRectZero style:UITableViewStylePlain];
     self.tableViewOfPausedCalls.backgroundColor = [UIColor blackColor];
     self.tableViewOfPausedCalls.delegate = self;
     self.tableViewOfPausedCalls.dataSource = self;
-    self.tableViewOfPausedCalls.hidden = YES;
     self.tableViewOfPausedCalls.separatorStyle = UITableViewCellSeparatorStyleNone;
     [self.tableViewOfPausedCalls registerClass:[PauseCallTableViewCell class] forCellReuseIdentifier:[PauseCallTableViewCell reuseIdentifier]];
 
-    [self.view addSubview:self.tableViewOfPausedCalls];
+    [self.tapOutsideTableViewButton addSubview:self.tableViewOfPausedCalls];
```
dvor: Not button, but `pauseCallsContainer`

Antidote/ActiveCallViewController.m
```diff
@@ -141,15 +144,28 @@ - (void)createCallMenuButton
...
+
+    self.tapOutsideTableViewButton = [UIButton new];
+    [self.tapOutsideTableViewButton addTarget:self
+                                       action:@selector(hideTableViewOfPausedCalls)
+                             forControlEvents:UIControlEventTouchUpInside];
+    [self.pauseCallsContainer addSubview:self.tapOutsideTableViewButton];
+
+    self.pauseCallsContainer.hidden = YES;
+
+    [self.view addSubview:self.pauseCallsContainer];
```
dvor: Can you please move this to `pauseCallsContainer` creation?

Antidote/ActiveCallViewController.m
```diff
@@ -223,10 +239,20 @@ - (void)installConstraints
         make.top.equalTo(self.topViewContainer.topMargin);
     }];
 
-    [self.tableViewOfPausedCalls makeConstraints:^(MASConstraintMaker *make) {
+    [self.pauseCallsContainer makeConstraints:^(MASConstraintMaker *make) {
         make.top.equalTo(self.topViewContainer.bottom);
         make.width.equalTo(self.view);
-        make.bottom.equalTo(self.view.bottom).with.offset(-kTableViewBottomOffSet);
+        make.bottom.equalTo(self.view);
```
dvor: Hm, why you want to place this view under `topViewContainer`? Top container is "outside" of tableView too. I think you can do `make.edges.equalTo(self.view);` here

Chuongv: That's because the tableView wouldn't know where to put the top at since it won't share a common superview with `self.topViewContainer`

dvor: Okay then.

Antidote/ActiveCallViewController.m
```diff
@@ -223,12 +242,25 @@ - (void)installConstraints
         make.top.equalTo(self.topViewContainer.topMargin);
     }];
 
-    [self.tableViewOfPausedCalls makeConstraints:^(MASConstraintMaker *make) {
+    [self.pauseCallsContainer makeConstraints:^(MASConstraintMaker *make) {
         make.top.equalTo(self.topViewContainer.bottom);
         make.width.equalTo(self.view);
```
dvor: I think that it would be better to set `left` and `right` instead of `width`. With `width` we don't have any horizontal constraints and 0 is used by default for `left`, which isn't that convenient.

Antidote/ActiveCallViewController.m
```diff
@@ -223,12 +242,25 @@ - (void)installConstraints
...
-        make.bottom.equalTo(self.view.bottom).with.offset(-kTableViewBottomOffSet);
+        make.bottom.equalTo(self.view);
     }];
 
+    [self.tableViewOfPausedCalls makeConstraints:^(MASConstraintMaker *make) {
+        make.top.equalTo(self.pauseCallsContainer);
+        make.width.equalTo(self.pauseCallsContainer);
```
dvor: Same here, `left` and `right` instead of `width`.

Antidote/ActiveCallViewController.m
```diff
@@ -223,12 +242,25 @@ - (void)installConstraints
...
+        make.bottom.equalTo(self.pauseCallsContainer).with.offset(-kTableViewBottomOffSet);
+    }];
+
+    [self.tapOutsideTableViewButton makeConstraints:^(MASConstraintMaker *make) {
+        make.top.equalTo(self.tableViewOfPausedCalls);
+        make.width.equalTo(self.pauseCallsContainer);
```
dvor: Same here

Antidote/ActiveCallViewController.m
dvor: Missed this one

PR #51, opened 2015-07-21
Some UI improvements

Chuongv:
Some slight UI improvements.

### DIFF
--- a/Antidote/AppearanceManager.m
+++ b/Antidote/AppearanceManager.m
@@ -221,6 +221,11 @@
     }
 }
 
+- (UIColor *)callRedColor
+{
+    return [UIColor colorWithRed:0.86 green:0.24 blue:0.24 alpha:1.0];
+}
+
 #pragma mark -  Private
 
 - (void)setColorscheme:(AppearanceManagerColorscheme)colorscheme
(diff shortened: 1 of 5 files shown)


Antidote/PauseCallTableViewCell.m
```diff
@@ -82,8 +83,9 @@ - (void)installConstraints
 {
     [self.endCallButton makeConstraints:^(MASConstraintMaker *make) {
         make.right.equalTo(self.contentView.rightMargin);
-        make.height.equalTo(self.contentView);
+        make.height.equalTo(kEndCallButtonWidth);
         make.width.equalTo(kEndCallButtonWidth);
```
dvor: This is not `Width` anymore, but `Size`. Also you can replace these two with `make.size.equalTo(kEndCallButtonSize);`

Antidote/AppearanceManager.m
```diff
@@ -221,6 +221,11 @@ - (UIColor *)bubbleIncomingColorForScheme:(AppearanceManagerColorscheme)scheme
     }
 }
 
+- (UIColor *)callRedColor
+{
+    return [UIColor colorWithRed:0.86 green:0.24 blue:0.24 alpha:1.0];
+}
```
dvor: This is good idea to move colors here. You can move other call related colors too (except for standard colors like `clear`, etc)

PR #53, opened 2015-07-22
Improve movement switching between calls.

Chuongv:
Summary:

Supports case when user hangs up a call while an incoming call is in place.
Bug Fix: No longer switches caller's name when pausing a current call when there is another pause call.
Bug Fix: Logic error when declining a call when there is already an incoming pending call.
Also some refactoring for calls manager.
For `currentCall`, I replaced `OCTCall` with its RealmSafeObject since it wanted to hold onto the primary key to see if the `currentCall` was removed. Otherwise I would get an exception that the object was invalidated/removed.


Antidote/CallsManager.m
```diff
@@ -451,13 +465,33 @@ - (void)switchViewControllerForCall:(OCTCall *)call
...
-            ([self.allPausedCallsController.fetchedObjects count] != 0));
+    OCTCall *call = [self.allPausedCallsController.fetchedObjects firstObject];
+
+    if (! call) {
+        call = [self.allCallsController.fetchedObjects firstObject];
+    }
+
+    // workaround for deadlock in objcTox
+    // https://github.com/Antidote-for-Tox/objcTox/issues/51
+    [self performSelector:@selector(switchViewControllerForCall:) withObject:call afterDelay:0];
```
dvor: I think you should return if there is no call at all

Antidote/CallsManager.m
```diff
@@ -140,17 +142,17 @@ - (void) controller:(RBQFetchedResultsController *)controller
...
 
-            if ([self onlyPauseCallsLeft]) {
-                OCTCall *call = [self.allPausedCallsController.fetchedObjects firstObject];
+            if (controller == self.allCallsController) {
 
-                // workaround for deadlock in objcTox
-                // https://github.com/Antidote-for-Tox/objcTox/issues/51
-                [self performSelector:@selector(switchViewControllerForCall:) withObject:call afterDelay:0];
+                if ([anObject.primaryKeyValue isEqualToString:self.currentCall.primaryKeyValue]) {
+                    [self replaceCurrentCallThatWasRemoved];
```
dvor: Having complete sentence as method name isn't that readable ("Replace current call that was removed"). I think it would be better to call this method `didRemoveCurrentCall`. I think it would be clear enough what this method means. And in future there may be some other logic than replacement the call.

Antidote/CallsManager.m
```diff
@@ -421,7 +435,7 @@ - (void)switchViewControllerForCall:(OCTCall *)call
 {
     AALogVerbose(@"%@", call);
 
-    if ([self.currentCallViewController isKindOfClass:[ActiveCallViewController class]]) {
+    if ([self isCurrentViewControllerReusableForCurrentCall:call]) {
```
dvor: This naming is strange. You check if there is reusable controller and then assume that it always is ActiveCallViewController

Antidote/CallsManager.m
```diff
@@ -451,13 +465,33 @@ - (void)switchViewControllerForCall:(OCTCall *)call
...
+    }
+
+    // workaround for deadlock in objcTox
+    // https://github.com/Antidote-for-Tox/objcTox/issues/51
+    [self performSelector:@selector(switchViewControllerForCall:) withObject:call afterDelay:0];
+}
+
+- (BOOL)isCurrentViewControllerReusableForCurrentCall:(OCTCall *)call
+{
+    if (! ((call.status == OCTCallStatusActive) || (call.status == OCTCallStatusPaused))) {
```
dvor: Hm, can this case happen? When
- `currentCallViewController` is ActiveCallViewContoller

and
- call isn't active or paused?

Antidote/CallsManager.m
Chuongv: If there are no calls from the pause calls, we want to grab a call from the others, which would be incoming calls. 

For example, Alice is in active call with Bob, Carol makes an incoming call. Alice ends call with Bob. In this case, there are no paused calls, so this method will take the incoming one.

Antidote/CallsManager.m
Chuongv: Yea it is. That's because ActiveCallViewController's are the only reusable ones in this case if it's in either a paused or active state. At the time, I was wondering if I should of done the check in the "if" clause or move it to another method. I chose the latter.

Is there a better naming I could of done?

Antidote/CallsManager.m
Chuongv: No that should never happen.

Antidote/CallsManager.m
Chuongv: Otherwise the other option:
`if ([self.currentCallViewController isKindOfClass:[ActiveCallViewController class] && (call.status == OCTCallStatusActive || call.status == OCTCallStatusPaused)) `

Which is quite messy

Antidote/CallsManager.m
dvor: Yeah, I understand that. What I meant was to check once again if there is no call at all before calling `performSelector...`. In this case CallsManager should be dead, but it would nice to have safer code.

```
OCTCall *call = [self.allPausedCallsController.fetchedObjects firstObject] ?:
                [self.allCallsController.fetchedObjects firstObject];

if (! call) {
    return;
}

[self performSelector:...];
```

Antidote/CallsManager.m
dvor: Then there is no need for this method. We can simply check if current controller is ActiveCallViewContoller. If it is active, it means that call is active or paused.

If there is no need in this method then there will be no problem with it strange naming. :-)

Or am I wrong?

Antidote/CallsManager.m
Chuongv: Ahhh I see. Good point :smile:

Antidote/CallsManager.m
Chuongv: Call A is active in progress. 
Call B is incoming.

User decides to end Call A, while ignoring Call B's incoming notification.

This is when it is necessary for us to check if the left over calls are, in which case is Call B, can be used with the current view controller.

dvor: Yeah, I see that now.

Antidote/CallsManager.m
dvor: When you'll remove `OCTCallStatusPaused` from objcTox it will look quite nice:
```
    if ((call.status == OCTCallStatusActive) &&
        [self.currentCallViewController isKindOfClass:[ActiveCallViewController class]]) {
    ...
```
So I think you can remove `isCurrentViewControllerReusableForCurrentCall` method after updating objcTox.

Chuongv: Yea I agree :+1:

Chuongv:
Will reopen once the refactor of objcTox pause states are done

PR #81, opened 2015-07-26
Feature: Add Proximity sensor while in call

Chuongv:
Screen will turn off when phone is placed up to the ear while in
a call.

### DIFF
--- a/Antidote/CallsManager.m
+++ b/Antidote/CallsManager.m
@@ -463,7 +463,8 @@
     AALogVerbose(@"%@", call);
 
     UIDevice *currentDevice = [UIDevice currentDevice];
-    currentDevice.proximityMonitoringEnabled = (call.status == OCTCallStatusActive);
+    currentDevice.proximityMonitoringEnabled = (call.status == OCTCallStatusActive ||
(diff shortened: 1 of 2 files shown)


Antidote/CallsManager.m
```diff
@@ -437,6 +441,9 @@ - (void)switchViewControllerForCall:(OCTCall *)call
 {
     AALogVerbose(@"%@", call);
 
+    UIDevice *currentDevice = [UIDevice currentDevice];
+    currentDevice.proximityMonitoringEnabled = (call.status == OCTCallStatusActive);
```
dvor: It should be enabled for dialing screen as well. Also if would be nice to enable proximityMonitoring only when loud speaker is off and user uses ear speaker.

dvor:
Conflicts. Also will you add proximity to Dialing controller?

dvor:
> Also will you add proximity to Dialing controller?

I've missed that it is already there

```
+    currentDevice.proximityMonitoringEnabled = (call.status == OCTCallStatusActive ||
+                                                call.status == OCTCallStatusDialing);
```

PR #85, opened 2015-07-27
Change mic disable to fill with white if selected

Chuongv:
Change it to be similar as the speaker button. No more changing image pictures.


Chuongv:
Closing, going to move this to another PR

Issue #86, opened 2015-07-28
Pause view screen


Chuongv:
When a friend pauses a call and user is at that call screen, the user should be presented a screen showing that the friend has paused the call.

PR #88, opened 2015-07-28
Feature: Pause View Screen #86

Chuongv:
This is for https://github.com/Antidote-for-Tox/Antidote/issues/86

Rather than creating a whole new view controller, I thought it would make more sense to use ActiveCallViewController since the call is considered `OCTCallActive`. I didn't add the "please wait..." label above the end call button. Do you think it's necessary? If so I need to figure out a way to juggle displaying the incoming call and please wait at the same time.

Here is a screen shot:
screen shot 2015-07-27 at 10 31 51 pm

### DIFF
--- a/Antidote/ActiveCallViewController.m
+++ b/Antidote/ActiveCallViewController.m
@@ -217,8 +217,7 @@ static const CGFloat kAvatarDiameter = 180.0;
     [self.view addSubview:self.friendAvatar];
 
     [self.friendAvatar makeConstraints:^(MASConstraintMaker *make) {
-        make.centerX.equalTo(self.view);
-        make.centerY.equalTo(self.view);
+        make.center.equalTo(self.view);
     }];
 }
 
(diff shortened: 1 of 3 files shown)


dvor:
> I didn't add the "please wait..." label above the end call button.

I think it is okay without this button.

Antidote/ActiveCallViewController.m
```diff
@@ -197,6 +202,27 @@ - (void)createBadgeViews
...
+                                     textColor:[UIColor whiteColor]
+                               backgroundColor:[UIColor clearColor]];
+
+    self.friendAvatar = [[UIImageView alloc] initWithImage:image];
+
+    [self.view addSubview:self.friendAvatar];
+
+    [self.friendAvatar makeConstraints:^(MASConstraintMaker *make) {
+        make.centerX.equalTo(self.view);
+        make.centerY.equalTo(self.view);
```
dvor: You can just use `center` instead of `centerX` and `centerY`. However it isn't critical.

Antidote/ActiveCallViewController.m
```diff
@@ -378,6 +415,8 @@ - (void)updateTimerLabel
     self.subLabel.text = [NSString stringFromTimeInterval:self.callDuration];
 
     [self.subLabel setNeedsDisplay];
+
+    [self removeFriendAvatar];
```
dvor: I think it would be better to rename method to `hideCallPausedByFriendIfNeeded`.

PR #103, opened 2015-08-11
Support video for Antidote

Chuongv:
Still pretty a lot to work on. But if you have any advice or suggestions, please let me know. 

Add CompactControlsView - Call controls shown horizontally
CallControlsView has a type that determines if it shows the
compacted control view or the normal view.


Antidote/ActiveCallViewController.m
```diff
@@ -234,6 +233,37 @@ - (void)installConstraints
...
+{
+    self.callControlsView.type = CallsControlsViewTypeCompact;
+    self.endCallButton.hidden = YES;
+
+    [self.callControlsView remakeConstraints:^(MASConstraintMaker *make) {
+        make.bottom.equalTo(self.view);
+        make.left.equalTo(self.view);
+        make.right.equalTo(self.view);
+    }];
+}
```
dvor: Hm, I'm wondering, maybe it would be better to 
- make `callControlsView` same size as controller's view
- make it first view in hierarchy
- make it transparent and passing gestures
- layout expendView at center, compactView at the bottom

What do you think about that? That way there will be no need in remaking constraints, you would need just to change `type` property.

Antidote/ActiveCallViewController.m
```diff
@@ -319,6 +349,71 @@ - (void)showCallPausedByFriend
...
+            make.bottom.equalTo(self.endCallButton.top).with.offset(-kIndent);
+            make.top.equalTo(self.topViewContainer.bottom).with.offset(kIndent);
+            make.left.equalTo(self.view);
+            make.right.equalTo(self.view);
+        }];
+
+        [self updateCallsControlViewCondensed];
+    }
+    else {
+        [self.videoView removeFromSuperview];
```
dvor: If there is no `self.videoView` there is nothing to remove from superview. You have to

```objc
- (void)setVideoView:(UIView *)videoView
{
    // sanity check
    if (_videoView == videoView) {
        return;
    }

    // check if there was old feed viev
    if (_videoView) {
        [_viewView removeFromSuperview];
        ...
    }

    // only then reassign it
    _videoView = videoView;

    // and if there is a new one, do stuff you do with it
    if (! videoView) {
        return;
    }
    
    // do stuff
    ...
}
```

Antidote/ActiveCallViewController.m
```diff
@@ -319,6 +349,71 @@ - (void)showCallPausedByFriend
...
+            [self updateCallsControlViewExpand];
+        }
+    }
+}
+
+- (void)setPreviewLayer:(CALayer *)previewLayer
+{
+    _previewLayer = previewLayer;
+
+    if (! self.previewLayer) {
```
dvor: Same stuff here

Antidote/CallControlsView.h
```diff
@@ -15,11 +15,31 @@
...
 
+typedef NS_ENUM(NSUInteger, CallsControlsViewType) {
+    /**
+     * Use this view for audio controls.
+     */
+    CallsControlsViewTypeExpand,
+
+    /**
+     * Use this view when showing video.
+     * Controls are set up horizontally |-video-mic-speaker-endCall-|
```
dvor: I think we want to use `compact` mode for horizontal audio as well. Comment is wrong then.

Antidote/ActiveCallViewController.m
Chuongv: Ohh that would be a good idea. I'll give that a try

Antidote/ActiveCallViewController.m
```diff
@@ -232,8 +232,28 @@ - (void)installConstraints
...
     }];
+
+    [self.callControlsView makeConstraints:^(MASConstraintMaker *make) {
+        make.edges.equalTo(self.view);
+    }];
 }
 
+- (void)viewDidLayoutSubviews
+{
+    [self adjustPreviewLayer];
```
dvor: It is better to call `[super viewDidLayoutSubviews]`

Antidote/ActiveCallViewController.m
```diff
@@ -232,8 +232,28 @@ - (void)installConstraints
...
+- (void)updateCallsControlViewExpand
+{
+    self.callControlsView.type = CallsControlsViewTypeExpand;
+    self.endCallButton.hidden = NO;
+}
+
+- (void)updateCallsControlViewCondensed
+{
+    self.callControlsView.type = CallsControlsViewTypeCompact;
+    self.endCallButton.hidden = YES;
```
dvor: Maybe move `endCallButton` to `callControlsView` as well?

Antidote/ActiveCallViewController.m
```diff
@@ -319,6 +339,84 @@ - (void)showCallPausedByFriend
...
+
+        return;
+    }
+
+
+    if (self.previewView) {
+        [self.view insertSubview:self.videoView belowSubview:self.previewView];
+    }
+    else {
+        [self.view addSubview:self.videoView];
```
dvor: This way `videoView` will be above `callControlsView`. I think that you can add transparent `videoContainer` at start, and add `videoView` and `previewView` there.

dvor: Also a nice feature to have - on tapping on `videoContainer` hide/show all controls (i.e. enter full screen mode)

Antidote/ActiveCallViewController.m
```diff
@@ -319,6 +339,84 @@ - (void)showCallPausedByFriend
...
+    }
+
+    [self.videoView makeConstraints:^(MASConstraintMaker *make) {
+        make.bottom.equalTo(self.endCallButton.top).with.offset(-kIndent);
+        make.top.equalTo(self.topViewContainer.bottom).with.offset(kIndent);
+        make.left.equalTo(self.view);
+        make.right.equalTo(self.view);
+    }];
```
dvor: Will `videoView` be resized properly? Or it will have wrong proportions (fix for that can go to next pull requests)

Antidote/ActiveCallViewController.m
```diff
@@ -319,6 +339,84 @@ - (void)showCallPausedByFriend
...
+
+    if (_previewLayer) {
+        [_previewLayer removeFromSuperlayer];
+    }
+
+    _previewLayer = previewLayer;
+
+    if (! _previewLayer) {
+        [self.callControlsView setVideoButtonSelected:NO];
+        return;
```
dvor: I think here you have to check if there is `videoView` and change type to `expand` if there isn't.

Actually I'm thinking that you can replace `updateCallsControlView{Condensed,Expand}` method with one `updateCallsControlViewType`, that will check if there is preview and viewView and change type appropriately. You can call this method in all needed places.

Antidote/ActiveCallViewController.m
```diff
@@ -319,6 +339,84 @@ - (void)showCallPausedByFriend
...
+            make.centerY.equalTo(self.videoView.bottom);
+        }
+        else {
+            make.centerY.equalTo(self.view);
+        }
+        make.right.equalTo(self.view);
+        make.size.equalTo(kVideoPreviewSize);
+    }];
+
+    self.previewLayer.frame = self.previewView.bounds;
```
dvor: It is better to call existing method `adjustPreviewLayer`

Antidote/ActiveCallViewController.m
Chuongv: Yea moving the videoView and preview stuff to a separate class right now.

Antidote/CallControlsView.m
```diff
@@ -8,18 +8,23 @@
...
+static const CGFloat k3ButtonGap = 30.0;
 
-@interface CallControlsView ()
+@interface CallControlsView () <CompactControlsViewDelegate>
 
 @property (strong, nonatomic) UIButton *videoButton;
 @property (strong, nonatomic) UIButton *microphoneButton;
 @property (strong, nonatomic) UIButton *speakerButton;
 @property (strong, nonatomic) UIButton *resumeButton;
+@property (strong, nonatomic) UIView *expandedView;
```
dvor: Maybe move buttons to expandedView?

Antidote/ActiveCallViewController.m
Chuongv: Actually if I do this, I won't have a frame of reference of where to the pin the bottom of the video view container since now the compact view won't share a common superview with the video container.

![image of videoView]
(http://i.snag.gy/lwCtQ.jpg)

I'm thinking about moving the compact view out of CallControlsView(rename this to expanded).

dvor: Well yeah, in fact you can add two views to controller - expanded and compact. And then hide/show them in update method.

Antidote/CallControlsView.m
```diff
@@ -91,28 +113,44 @@ - (void)installConstraints
...
+        make.bottom.equalTo(self);
+        make.left.equalTo(self);
+        make.right.equalTo(self);
     }];
 }
 
+- (void)layoutSubviews
+{
+    [super layoutSubviews];
+    [self layoutIfNeeded];
```
dvor: Subviews are already layout-ed here. Why do you need to call `layoutIfNeeded` again?

Antidote/CallControlsView.m
```diff
@@ -130,11 +168,57 @@ - (void)resumeButtonPressed
...
+}
+
+#pragma mark - Public
+
+- (void)setType:(CallsControlsViewType)type
+{
+    _type = type;
+
+    self.compactView.hidden = (type == CallsControlsViewTypeExpand);
+    self.expandedView.hidden = (type == CallsControlsViewTypeCompact);
```
dvor: Despite the fact that compact view will move to controller, it would be better to do
```
self.compactView.hidden = (type != CallsControlsViewTypeCompact);
self.expandedView.hidden = (type != CallsControlsViewTypeExpand);
```

That way there will be no errors if we'll add more types.

Antidote/CallsManager.m
```diff
@@ -222,6 +222,23 @@ - (void)updateCurrentCallInterface:(OCTCall *)call
         else {
             activeVC.callDuration = call.callDuration;
         }
+
+        if (call.friendSendingVideo && ! activeVC.videoView) {
```
dvor: No need for this check `! activeVC.videoView`. Controller will handle case of adding same `videoView`

Antidote/CallsManager.m
```diff
@@ -222,6 +222,23 @@ - (void)updateCurrentCallInterface:(OCTCall *)call
...
+            AALogVerbose(@"provide video feed to video view");
+            activeVC.videoView = [self.manager videoFeed];
+        }
+        else if (! call.friendSendingVideo && ! call.videoIsEnabled) {
```
dvor: And if you'll change first if there will be no need for `! call.friendSendingVideo` check here.

Also, why do we rely on two different properties `friendSendingVideo` and `videoIsEnabled`. Should there be just one of them?

Antidote/CallsManager.m
```diff
@@ -222,6 +222,23 @@ - (void)updateCurrentCallInterface:(OCTCall *)call
...
+            activeVC.videoView = nil;
+        }
+
+        if (call.videoIsEnabled && ! activeVC.previewLayer) {
```
dvor: No need for `! activeVC.previewLayer` check as well

Antidote/ActiveCallViewController.m
Chuongv: Yep that's my plan :smile:

Antidote/CallsManager.m
```diff
@@ -441,6 +470,19 @@ - (AbstractCallViewController *)createViewControllerForCall:(OCTCall *)call
...
+                activeVC.videoView = [self.manager videoFeed];
+            }
+
+            if (call.videoIsEnabled) {
+                __weak ActiveCallViewController *weakVC = activeVC;
+                [self.manager getVideoCallPreview:^(CALayer *layer) {
+                    ActiveCallViewController *strongVC = weakVC;
+                    strongVC.previewLayer = layer;
+                }];
+            }
```
dvor: Hm, some code duplication with `updateCurrentCallInterface:` method. It would be better to avoid it

Antidote/CallsManager.m
```diff
@@ -483,6 +525,18 @@ - (void)switchViewControllerForCall:(OCTCall *)call
...
+            activeVC.videoView = [self.manager videoFeed];
+        }
+
+        if (call.videoIsEnabled) {
+            __weak ActiveCallViewController *weakVC = activeVC;
+            [self.manager getVideoCallPreview:^(CALayer *layer) {
+                ActiveCallViewController *strongVC = weakVC;
+                strongVC.previewLayer = layer;
+            }];
+        }
```
dvor: More code duplication

Antidote/ActiveCallViewController.m
dvor: Cool :+1:

Antidote/CallControlsView.m
Chuongv: Hmm yea, I think I was following some stackoverflow answer about updating rounded corners. I will remove and test to see if it works without.

Antidote/CompactControlsView.h
```diff
@@ -0,0 +1,47 @@
...
+
+@class CompactControlsView;
+
+@protocol CompactControlsViewDelegate <NSObject>
+
+- (void)compactControlsMicButtonPressed:(CompactControlsView *)compactControlsView;
+- (void)compactControlsSpeakerButtonPressed:(CompactControlsView *)compactControlsView;
+- (void)compactControlsResumeButtonPressed:(CompactControlsView *)compactControlsView;
+- (void)compactControlsVideoButtonPressed:(CompactControlsView *)compactControlsView;
+- (void)compactControlsEndCallButtonPressed:(CompactControlsView *)compactControlsView;
```
dvor: Compact and expanded views can have same protocol for their interface and same protocol for delegate. Pseudo code:

```
@protocol CallControlsViewProtocol
@property (assign, nonatomic) BOOL videoButtonSelected; 
...

@protocol CallControlsViewDelegate
- (void)callControlsViewMicButtonPressed:(id<CallControlsViewProtocol>)view;
...

@interface CompactControlsView <CallControlsViewProtocol>

@property (weak, nonatomic) id<CallControlsViewDelegate> delegate;

@end

// same for expand view
```

Antidote/CompactControlsView.m
```diff
@@ -0,0 +1,194 @@
...
+    [self.endCallButton makeConstraints:^(MASConstraintMaker *make) {
+        make.left.equalTo(self.speakerButton.right).with.offset(kSpacing);
+        make.right.equalTo(self).with.offset(-kSpacing);
+        make.centerY.equalTo(self.videoButton);
+        make.size.equalTo(self.videoButton);
+    }];
+
+    [self makeConstraints:^(MASConstraintMaker *make) {
+        make.top.equalTo(self.videoButton);
+        make.bottom.equalTo(self.videoButton);
```
dvor: You can move this to `[self.videoButton makeConstraints:..` on top

Antidote/CallsManager.m
Chuongv: Sorry I missed this. `friendSendingVideo` and `videoIsEnabled` is both needed to know if we need to update the UI for outgoing video versus just incoming video `friendSendingVideo`.

Chuongv:
Okay to review, though I am still trying to figure out how to present the preview layer without locking up the main queue. Everything seems to work great for vertical video. Haven't tested this with a 3rd party incoming call yet, so more refactoring with auto-layout will be needed with that.

Antidote/ActiveCallViewController.h
```diff
@@ -64,11 +65,26 @@
...
 
 /**
+ * Video view of friend
+ */
+@property (strong, nonatomic) UIView *videoView;
+
+/**
+ * Preview layer of video sending
+ */
+@property (strong, nonatomic) CALayer *previewLayer;
```
dvor: You can make this properties private and add public methods instead
```
- (void)provideVideoView:(UIView *)view;
- (void)providePreviewLayer:(CALayer *)layer;
```

I think that this would make API more explicit. You actually don't read this properties out of controller.

Antidote/ActiveCallViewController.m
```diff
@@ -393,7 +442,14 @@ - (void)hideCallPausedByFriendIfNeeded
...
 
-    self.callControlsView.hidden = NO;
+    self.expandedControlsView.hidden = NO;
+}
+
+- (void)switchToVideoViewIfNeeded
+{
+    self.videoContainerView.hidden = ! (self.videoContainerView.videoView || self.videoContainerView.previewLayer);
+    self.expandedControlsView.hidden = (self.videoContainerView.videoView || self.videoContainerView.previewLayer);
+    self.compactControlsView.hidden = ! (self.videoContainerView.videoView || self.videoContainerView.previewLayer);
```
dvor:
```objc
BOOL videoVisible = (self.videoContainerView.videoView || self.videoContainerView.previewLayer);

self.videoContainerView.hidden = ! videoVisible;
self.expandedControlsView.hidden = videoVisible;
self.compactControlsView.hidden = ! videoVisible;
```

Antidote/CallsManager.m
```diff
@@ -50,7 +50,7 @@ @implementation CallsManager
 - (instancetype)init
 {
     AALogVerbose();
-    self = [super self];
+    self = [super init];
```
dvor: Wow, what was thas? :grinning:

Antidote/CallsManager.m
```diff
@@ -222,6 +222,36 @@ - (void)updateCurrentCallInterface:(OCTCall *)call
...
+            // Even with the below, getting or removing the the preview layer
+            // is locking up the main queue.
+            //            CALayer *dummyLayer = [CALayer new];
+            //            activeVC.previewLayer = dummyLayer;
+            //
+            //            __weak ActiveCallViewController *weakVC = activeVC;
+            //            [self.manager getVideoCallPreview:^(CALayer *layer) {
+            //                ActiveCallViewController *strongVC = weakVC;
+            //                strongVC.previewLayer = layer;
+            //            }];
```
dvor: I think you should have `BOOL` property on call to distinguish if preview is enabled or not.

Antidote/ActiveCallViewController.h
Chuongv: In CallsManager I actually do read the properties from the controller. I had it originally methods like the above, but then CallsManager was continuously invoking methods from `OCTVideoEngine` which overloaded the thread. Do you know if there's a better way?

```
        if (call.friendSendingVideo && ! activeVC.videoView) {
            activeVC.videoView = [self.manager videoFeed];
        }

        if (! call.friendSendingVideo && activeVC.videoView) {
            activeVC.videoView = nil;
        }

        if (! call.videoIsEnabled && activeVC.previewLayer) {
            activeVC.previewLayer = nil;
            activeVC.videoButtonSelected = NO;
        }
```

Antidote/ActiveCallViewController.m
Chuongv: Ah thanks, this is much cleaner :smile:

Antidote/CallsManager.m
Chuongv: It's like the 3rd time I done that. Surprisingly not sure why it didn't cause any problems before.

Antidote/CallsManager.m
Chuongv: Yea maybe I should do the same as well for `videoView`? That way we can go back to using methods like you suggested on `ActiveCallViewController` instead of properties.

dvor: Yes, I think it should work

Antidote/ExpandedControlsView.m
```diff
@@ -0,0 +1,290 @@
...
+}
+
+- (BOOL)resumeButtonHidden
+{
+    return self.resumeButton.hidden;
+}
+
+- (void)setVideoButtonSelected:(BOOL)selected
+{
+    NSAssert( YES, @"Video button should never be set here");
```
dvor: `NSAssert(YES)` won't fail. :grinning: You have to do `NSAssert(NO, ...)`

Antidote/ExpandedControlsView.m
```diff
@@ -0,0 +1,290 @@
...
+}
+
+- (BOOL)videoButtonSelected
+{
+    NSAssert(YES, @"Video button selected should not be checked here");
```
dvor: Same here

Antidote/VideoAndPreviewView.m
```diff
@@ -0,0 +1,139 @@
...
+    self.previewView.backgroundColor = [UIColor blackColor];
+    self.previewView.hidden = YES;
+
+    [self addSubview:self.previewView];
+}
+
+- (void)layoutSubviews
+{
+    [super layoutSubviews];
+    [self layoutIfNeeded]; // is this needed?
```
dvor: So is it needed?

Antidote/VideoAndPreviewView.m
```diff
@@ -0,0 +1,139 @@
...
+        make.bottom.equalTo(self.previewView);
+        make.left.equalTo(self);
+        make.right.equalTo(self);
+        make.top.equalTo(self);
+    }];
+}
+
+- (void)setPreviewLayer:(CALayer *)previewLayer
+{
+    dispatch_async(dispatch_get_main_queue(), ^{
```
dvor: Hm, controller definitely shouldn't do that. `previewLayer` should be delivered on main thread already (from objcTox I think)

Chuongv: Yea you are right. Though setting it to nil is not on the main queue though. I have to redo this logic, since even removing the layer from the superlayer while on the main queue causes issues/slowness.

PR #109, opened 2015-08-14
Bug fix #106 preview layer removal slowness

### LINKED ISSUE #106: Removing an existing preview layer locks up the main queue
In `VideoAndPreviewView`, `[_previewLayer removeFromSuperlayer];` takes 5+ seconds to remove. It locks up the main queue and all operations of the app.

Chuongv:
Rather than remove the layer after video engine stops running, the
`VideoViewAndPreview` will hold onto it until the user exits out of video mode.

https://github.com/Antidote-for-Tox/Antidote/issues/106

### DIFF
--- a/Antidote/ActiveCallViewController.h
+++ b/Antidote/ActiveCallViewController.h
@@ -83,9 +83,9 @@
  * Since it takes a while for the preview layer to be loaded,
  * it is best to set this to YES to indicate that one is loaded and in progress
  * or that it is already present.
- * Set to YES if the preview view is being shown, otherwise NO.
+ * Set to YES before providing a preview layer.
  */
-@property (nonatomic, assign) BOOL previewViewIsShown;
+@property (nonatomic, assign) BOOL previewViewLoaded;
 
 /**
  * Create an incoming call view for friend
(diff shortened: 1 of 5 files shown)


Antidote/ActiveCallViewController.h
```diff
@@ -83,9 +83,9 @@
  * Since it takes a while for the preview layer to be loaded,
  * it is best to set this to YES to indicate that one is loaded and in progress
  * or that it is already present.
- * Set to YES if the preview view is being shown, otherwise NO.
+ * Set to YES before providing a preview layer.
  */
-@property (nonatomic, assign) BOOL previewViewIsShown;
+@property (nonatomic, assign) BOOL previewViewLoaded;
```
dvor: Hm, do we need this property on controller? Maybe you can move it to CallsManager?

Chuongv: Hmm. I think it has more relation to the ActiveCallViewController than the CallsManager since the ActiveCallViewController holds onto that previewview. If the ActiveViewController gets removed or replace, then we have to update it back to `NO` on the CallsManager.

dvor: Well, okay then. :-)

PR #112, opened 2015-08-15
Change button styles for compactViewControls

Chuongv:
Better look for the buttons. Though maybe I shouldn't change the alpha for the end call.
![compact buttons]
(http://i.snag.gy/IWkDN.jpg)


dvor:
I think they are okay. :-)

Issue #111, opened 2015-08-15
Rotation in video call causes a memory bad access error


Chuongv:
Every time the phone is rotated, there would be an exception thrown in `OCTVideoEngine`. Bad access errors to either the yPlane or uvPlane. I have a feeling it is related to too many views being rotated? This does not happen in `objcTox` demo

Chuongv:
Resolved by several fixes

PR #113, opened 2015-08-16
Set opaque to YES when video is shown

Chuongv:
Resolves https://github.com/Antidote-for-Tox/Antidote/issues/111

Though I might want to move video view controller to a separate view controller to since `ActiveCallViewController` is getting a little bit.


dvor:
It isn't that big for now. But you can try to move it out indeed.

Issue #134, opened 2015-08-20
Slight name cutoff at the bottom when in landscape


Chuongv:
See

![image cuttoff]
(https://cloud.githubusercontent.com/assets/5170710/9397554/be5a1bb4-4752-11e5-8d78-cc6ea78e76aa.PNG)

PR #137, opened 2015-08-21
Fix name but off when landscape

Chuongv:
https://github.com/Antidote-for-Tox/Antidote/issues/134


dvor:
Have you tested on different iPhones? 4s, 6, 6+?

Chuongv:
No not yet, good idea :smile:

Chuongv:
Tested for all iphones now

Issue #110, opened 2015-08-14
Fix autolayout for Dialing and Ringing view controllers


Chuongv:
They should support autolayout.

Chuongv:
Actually looking at skype, they support rotation. @dvor  what do you think?

dvor:
Well, it would be nice to support rotation. However it will take some work. I would suggest to leave it for now, lock orientation and polish thing that are already there (maybe investigate video quality issue as well).

Next Monday is already "pencils down" day.

Chuongv:
Okay, I change the issue's title to support this feature instead.

PR #141, opened 2015-08-23
Fix autolayout for dialing call view control

Chuongv:
For https://github.com/Antidote-for-Tox/Antidote/issues/110
Tested on all screen sizes and it looks good landscape/portrait :+1:

### DIFF
--- a/Antidote/DialingCallViewController.m
+++ b/Antidote/DialingCallViewController.m
@@ -39,11 +39,6 @@ static const CGFloat kEndCallButtonHeight = 45.0;
     [self installConstraints];
 }
 
-- (void)viewDidLayoutSubviews
-{
-    [self updateFriendAvatar];
-}
-
 - (void)installConstraints
 {
     [super installConstraints];
@@ -91,8 +86,6 @@ static const CGFloat kEndCallButtonHeight = 45.0;
 - (void)createFriendAvatar
 {
     self.friendAvatar = [UIImageView new];
-
-    [self updateFriendAvatar];
 
     [self.view addSubview:self.friendAvatar];
 }
--- a/Antidote/DialingCallViewController.m
+++ b/Antidote/DialingCallViewController.m
@@ -13,7 +13,7 @@
 #import "AppearanceManager.h"
 
 static const CGFloat kIndent = 50.0;
-static const CGFloat kAvatarDiameter = 180.0;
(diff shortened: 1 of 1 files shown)


Antidote/DialingCallViewController.m
```diff
@@ -84,8 +106,11 @@ - (void)updateFriendAvatar
 {
     AvatarsManager *avatars = [AppContext sharedContext].avatars;
 
+    CGFloat smallestSide = MIN(self.view.bounds.size.width, self.view.bounds.size.height);
+    CGFloat diameterOfImage = smallestSide / 2;
```
dvor: Hm... this value will always be the same. It doesn't matter how you rotate device - MIN(width, height) or MIN(height, width) will always give you same value :wink:

Chuongv: Oh yeaaa you are right. I don't need to update this every time I rotate then.

dvor:
>  5 minutes ago
> Chuongv pushed to chuongv/dialingControllerUpdate at Antidote-for-Tox/Antidote
>    @Chuongv 81cd9eb
>        Remove unnecessary avatar update 

---

> build 9
> Started	5 minutes ago
> Duration	03:24

Circleci is so fast, I'm really happy with it. :-)

Chuongv:
Wow that is quick

Chuongv:
Oh forgot one more thing with this PR

Antidote/DialingCallViewController.m
```diff
@@ -39,6 +39,28 @@ - (void)viewDidLoad
     [self installConstraints];
 }
 
+- (void)viewDidLayoutSubviews
+{
+    [self updateFriendAvatar];
+}
```
dvor: Now I'm thinking if we should call `updateFriendAvatar` method here. Afaik `viewDidLayoutSubviews` is called quite a lot of times. In case if we need to update it here it is better to add following check to `updateFriendAvatar` method

```
if (self.friendAvatar.frame.size == diameterOfImage) {
   return;
}
```

Chuongv: I removed it form `viewDidLayoutSubviews`. The only reason we would call `updateFriendAvatar` is if the name of the person changes or we are adding in a name. In this case I would like to recreate the image.

PR #159, opened 2015-09-12
Fix build by specifying commit in objcTox

Chuongv:
This will just get it to build for now. Master is pointing to tag objcTox 0.2.1 but since there is no objcTox tag for audio, I have to point it by commit for now temporarily. Once Master gets updated to latest objcTox, we can change this back to audio branch.


dvor:
Hm. It seems that I've broken audio branch somehow.

PR #206, opened 2016-08-01
Replace In-App-Notifications with LNNotificationsUI

sirnacnud:
This commit replaces the In-App-Notifications implementation with LNNotificationsUI.

Couple of things to note:

1. `LNNotificationsUI` mimics the iOS banner notifications, so the icon it displays in the notification is supposed to be an app icon, resulting in the notification looking poor if the`NotificationObject` image is specified. Because of this, the Antidote app icon is now displayed instead of the previous avatar image.

1. In-App-Notifications no longer wait until the notification is dismissed to display a new In-App-Notifications. `LNNotificationsUI` can display a new notification while the old one is still displayed.


dvor:
Thanks, I'll take a look.

Travis is broken for Antidote, so don't mind red build.

dvor:
Looks good, I've found only one issue - notifications aren't canceled when user goes to appropriate chat screen. 

Let's say user receives bunch of messages, taps notification with one of them, chat controller is pushed. After that it makes sense to remove all notifications that are related to opened chat - user will see all those messages anyway. Previously it was done by removing messages from queue in `banNotificationsForChat` method.

I think that we should remove only chat related notifications, not all of them. Imagine situation when user receives message A from Alice and message B from Bob. User taps on A, moves to chat with Alice. It still makes sense to display Bob's notification.

Is same behavior possible with LNNotificationsUI? Can you remove all notifications that are related to the particular chat?

sirnacnud:
`LNNotificationsUI` has support to clear pending notifications for a specific identifier, so I was able to implement the same behavior as before.

dvor:
Cool, thanks!

PR #207, opened 2016-08-03
Fix tests not compiling

sirnacnud:
ChatIncomingTextCellSnapshotTest and ChatOutgoingTextCellSnapshotTest needed to be updated to use ChatBaseTextCellModel, as the models they were using no longer existed.


dvor:
Thanks!

PR #209, opened 2016-08-05
Standardize ChatListCell and FriendListCell UI

sirnacnud:
I recently noticed the cells for the Contacts and Chats lists were sized different, making the UI seem off when switching between the two.  This change standardizes the UI elements in both ChatListCell and FriendListCell to have the same size and appearance.

In Particular:

- Avatars are now the same size
- Bottom labels are now the same size
- The right arrow is in the same position
- Chat message doesn't extend past the end of the date label

### DIFF
--- a/Antidote/ChatListCell.swift
+++ b/Antidote/ChatListCell.swift
@@ -19,7 +19,7 @@ class ChatListCell: BaseCell {
         static let MessageLabelHeight = 22.0
 
         static let NicknameToDateMinOffset = 5.0
-        static let DateToArrowOffset = 5.0
+        static let DateToArrowOffset = -5.0
 
         static let RightOffset = -7.0
         static let VerticalOffset = 3.0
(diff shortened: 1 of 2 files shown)


dvor:
Hi @sirnacnud,

Thanks for these pull request. Actually different size for contacts and chats cells was done on purpose. The idea is to make user intuitively know where he is at the moment.

I'm thinking to leave contacts cells small as it is and make chats cells bigger. Second line of the text can be added, something similar to what [Telegram has](https://s1.mzstatic.com/eu/r30/Purple4/v4/01/bf/64/01bf6491-20df-2c83-6bed-310da08f9828/screen568x568.jpeg).

If you like you can implement it. Closing this pull request for now.

sirnacnud:
I guess that's a fair argument, though the user can just look at the tab bar and see whether they are on Contacts or Chats. 

I do think having the right arrow in the same position on both cells is still a worth while change. I feel for FriendListCell it is too close to the right edge of the screen and on ChatListCell it should be vertically centered. For the standard UITableViewCell, the disclosure indicator is always vertically centered in the cell.

dvor:
With right arrow I was following iMessage design. They also have arrow on the top. Maybe it has slightly more space from the right border.

I think I can try to play with it and see if it will be better.

sirnacnud:
Ah, I see. I wonder why Apple chose to do that. I was basing my change off of WhatsApp, which has the arrow centered.

PR #224, opened 2016-09-21
Update README.md

Geremia:
clarified that email addresses (not email messages from Apple) should be sent to beta@antidote.im


dvor:
Thanks!

PR #297, opened 2017-04-12
Fix markdown formatting

nurupo:
[GitHub has changed their Markdown parser](https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown), which broke several previously working things, one of which is `#Heading`. It's now required to have a space after `#`, i.e. `# Heading` is the correct heading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 See [FAQ](FAQ/en.md).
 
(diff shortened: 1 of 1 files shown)


dvor:
Cool, thanks!

